PR #45, opened 2019-02-15
Change homepage call to action to go to install docs

jazzdan:
Not tied to this but I feel like the GitHub does not make a good step two. All the relevant information is below the fold on desktop and not visible on mobile.

### DIFF
--- a/src/index.md
+++ b/src/index.md
@@ -15,7 +15,7 @@ of your system, from building to deploying to logging to crashing.
 
 <div class="block u-marginTop2 u-marginBottom1_5">
 <div class="brandButton">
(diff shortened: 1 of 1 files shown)


src/index.md
```diff
@@ -15,7 +15,7 @@ of your system, from building to deploying to logging to crashing.
 
 <div class="block u-marginTop2 u-marginBottom1_5">
 <div class="brandButton">
-  <a class="imageLink" href="https://github.com/windmilleng/tilt">
+  <a class="imageLink" href="https://docs.tilt.build/install.html">
```
nicks: can you make this {{site.docsurl}} so that it works ok locally?

PR #60, opened 2019-02-26
api: filter_yaml docs

maiamcc:


api/api.py
```diff
@@ -137,6 +137,43 @@ def k8s_resource(name: str, yaml: Union[str, Blob] = "", image: Union[str, FastB
   """
   pass
 
+def filter_yaml(yaml: Union[str, List[str], LocalPath, Blob], labels={}, name='', kind='') -> Tuple[Blob, Blob]:
```
landism: you could add type annotations to `labels`, `name`, and `kind` instead of putting them in manually down below, to get consistent formatting

maiamcc:
screen shot 2019-02-26 at 2 49 48 pm

api/api.py
```diff
@@ -137,6 +137,43 @@ def k8s_resource(name: str, yaml: Union[str, Blob] = "", image: Union[str, FastB
...
+    # extract all YAMLs matching labels "app=foobar"
+    foobar_yaml, rest = filter_yaml('all.yaml', labels={'app': 'foobar'}
+    k8s_resource('foobar', yaml=foobar_yaml)
+
+    # extract YAMLs of kind "deployment" with metadata.name "baz"
+    baz_yaml, rest = filter_yaml(rest, name='baz', kind='deployment')
+    k8s_resource('baz', yaml=baz_yaml)
+
+    # apply the rest of the YAML as free-floating k8s entities
+    # (i.e., not attached to a Tilt resource)
```
landism: I'd worry a bit about a user reading this and thinking this means k8s_yaml is for applying unresourced entities, when I think we want the common case to be using k8s_yaml for applying resourced entities.

Maybe an improvement would be (and I think probably a good idea anyway?) to get a bit more explicit about the use case somewhere? something like this maybe? "for cases where you need to pull out just a bit of yaml to pass to `k8s_resource`, or when you have combined yaml configs and don't want to edit them, but also don't want to give *all* of them to Tilt")

PR #62, opened 2019-02-28
fb.add may take string (make this default example in docs) [ch1716]

maiamcc:
Hello @jazzdan, @landism,

Please review the following commits I made in branch maiamcc/add-takes-string:

c89db3f0f511e45f69417178339cb9f9c3a0f011 (2019-02-28 15:24:57 -0500)
fb.add may take string (make this default example in docs)

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -5,20 +5,19 @@ read_file('api/api.py')
 local('make api')
 
 k8s_yaml('serve.yaml')
-repo = local_git_repo('.')
 
 img = fast_build('gcr.io/windmill-public-containers/tilt-site',
                  'base.dockerfile',
                  'bundle exec jekyll serve --config _config.yml,_config-dev.yml')
-img.add(repo.path('src'), '/src/')
+img.add('./src', '/src/')
 img.run('bundle update', trigger=['src/Gemfile', 'src/Gemfile.lock'])
 img.hot_reload()
 
 img = fast_build('gcr.io/windmill-public-containers/docs-site',
                  'docs.dockerfile',
                  'bundle exec jekyll serve --config _config.yml,_config-dev.yml')
-img.add(repo.path('src'), '/src/')
-img.add(repo.path('docs'), '/docs/')
+img.add('./src', '/src/')
+img.add('./docs', '/docs/')
 img.run('bundle update', trigger=['src/Gemfile', 'src/Gemfile.lock', 'docs/Gemfile', 'docs/Gemfile.lock'])
 img.hot_reload()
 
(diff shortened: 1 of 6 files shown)


Tiltfile
```diff
@@ -5,20 +5,19 @@ read_file('api/api.py')
 local('make api')
 
 k8s_yaml('serve.yaml')
-repo = local_git_repo('.')
 
 img = fast_build('gcr.io/windmill-public-containers/tilt-site',
                  'base.dockerfile',
                  'bundle exec jekyll serve --config _config.yml,_config-dev.yml')
-img.add(repo.path('src'), '/src/')
+img.add('./src', '/src/')
```
jazzdan: Why do we need the `./`?

docs/tiltfile_concepts.md
```diff
@@ -12,6 +12,8 @@ Functions like `k8s_yaml` and `docker_build` register information. At the end of
 
 Because your Tiltfile is a program, you can configure it with familiar constructs like loops, functions, arrays, etc. This makes Tilt more extensible than a configuration that requires hard-coding all possible options up-front.
 
+Any relative paths in your Tiltfile are evaluated relative to the location of the Tiltfile. (Absolute paths are also fine!)
```
jazzdan: Should we encourage people to use absolute paths?

maiamcc: I don't think so -- seems like it adds more mental overhead, makes it harder to just scan a tiltfile and get what's happening, etc.

(Unless you're saying I should remove the last because it implicitly encourages abs paths and we DON'T want that?

Tiltfile
maiamcc: i mean, we don't -- imo it makes the relative-ness of the path more apparent, and also makes it more apparent that this arg is a path and not some other random thing (this is maybe not SO relevant in this example, but for instance i think in a call to
```
docker_build('gcr.io/stuff/thing', 'thing')
```
it's kind of ambiguous what that second arg is (is it maybe the name we assign to the tilt resource for this image? is it...??!) but a call to
```
docker_build('gcr.io/stuff/thing', './thing')
```
makes it v obvious that the second arg is a path, and then it's easier to reason about what it means

docs/tiltfile_concepts.md
jazzdan: Yeah that question is more what I'm asking :)

Absolute paths are annoying in shared code because someone else might have that file in a different path.

Tiltfile
jazzdan: I think @dbentley expressed a preference for leaving `./` out of paths, but for the record I also like being explicti here.

docs/tiltfile_concepts.md
maiamcc: Eh i don't have strong feelings. I can remove that parenthetical.

maiamcc:
okay, ready!

PR #68, opened 2019-03-05
api docs: add k8s_kind

landism:


api/api.py
```diff
@@ -205,3 +205,22 @@ def listdir(directory: str, recursive: bool = False) -> List[str]:
...
   pass
 
+def k8s_kind(kind: str, *, image_json_path: Union[str, List[str]]):
+  """Tells Tilt about a k8s kind. Primarily intended for defining where your CRD specifies image names.
+
+  (Note the `*` in the signature means `image_json_path` must be passed as a keyword, e.g., `image_json_path="{.spec.image}"`)
+
+  Example ::
+
+    # Fission has a CRD named "Environment"
```
jazzdan: Can you make this formatted with ```? (not sure if you can)

landism: That's [taken care of](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) by the `::`

image

PR #74, opened 2019-03-14
api docs: update reflecting regexes and api_version

landism:
`filter_yaml` and `k8s_kind` now take an `api_version` parameter, and also their selector params are (mostly) regexes

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -134,10 +134,10 @@ def filter_yaml(yaml: Union[str, List[str], Blob], labels: dict=None, name: str=
       must satisfy all of the specified label constraints, though they may have additional
       labels as well: see the `Kubernetes docs <https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/>`_
       for more info.)
-    name: Regexp specifying the ``metadata.name`` property of entities to match
-    namespace: Regexp specifying the ``metadata.namespace`` property of entities to match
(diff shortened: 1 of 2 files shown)


api/api.py
```diff
@@ -130,14 +130,14 @@ def filter_yaml(yaml: Union[str, List[str], Blob], labels: dict=None, name: str=
...
       must satisfy all of the specified label constraints, though they may have additional
       labels as well: see the `Kubernetes docs <https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/>`_
       for more info.)
-    name: (optional) the ``metadata.name`` property of entities to match
-    namespace: (optional) the ``metadata.namespace`` property of entities to match
-    kind: (optional) the kind of entities to match (e.g. "service", "deployment", etc.).
-      Case-insensitive. NOTE: doesn't support abbreviations like "svc", "deploy".
+    name: Regexp specifying the ``metadata.name`` property of entities to match
+    namespace: Regexp specifying the ``metadata.namespace`` property of entities to match
+    kind: Regexp specifying the kind of entities to match (e.g. "Service", "Deployment", etc.).
```
nicks: your change from service -> Service here made me wonder if these regexps should be case-insensitive. there are definitely k8s properties that don't even let you use upper-case to avoid case-sensitivity problems

PR #76, opened 2019-03-14
secret doc for nested fastbuild [ch1877]

maiamcc:
Hello @nicks, @landism,

Please review the following commits I made in branch maiamcc/nested-fastbuild:

090c3b8d72b166c6e81e04c1c46b42b94988cdf3 (2019-03-14 18:04:27 -0400)
secret doc for nested fastbuild


docs/nested_fastbuild.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Fast Build With Your Existing Dockerfile
+layout: docs
+---
+
+Want to take advantage of `fast_build` for lightning-fast live updates of your running containers, but don't want to redo your whole Dockerfile for it? Now you don't have to.
```
nicks: i would nix the 'lightning-fast', you don't need it

docs/nested_fastbuild.md
```diff
@@ -0,0 +1,36 @@
...
+
+You can read more in [Optimizing A Tiltfile](/fast_build.html), but in brief, when you make a call to `fast_build`, you specify a set of instructions that can be used to _both_:
+
+1. Build your image for the first time, and
```
nicks: fwiw, i've been putting newlines between list items

docs/nested_fastbuild.md
```diff
@@ -0,0 +1,36 @@
...
+```python
+img_name = 'myproject/server'
+(fast_build(img_name, 'Dockerfile.stripped', entrypoint='/go/bin/server')
+  .add('./server', '/go/src/myproject/server')
+  .run('go install myproject/server'))
+```
+
+Who wants to maintain two separate Dockerfiles, when part of the point of Tilt is to get your dev environment as close to your prod environment as possible?
+
+Well, now we have a solution. You can now specify steps for (1) and (2) separately: use your existing Dockerfile when building the image from scratch, and add specific files/run specific commands when updating in place.
```
nicks: i'm a little bit worried that like >50% of this doc is throat-clearing about the old stuff. why would the user care about what the old syntax was?  the reader won't be familiar with this, so it adds a lot of unnecessary work without increasing understanding

docs/nested_fastbuild.md
```diff
@@ -0,0 +1,36 @@
...
+
+```python
+img_name = 'myproject/server'
+
+# use your existing Dockerfile for image builds
+server = docker_build(img_name, ',/server')
+
+# specify how to update the image in place
+server.add('./server', '/go/src/myproject/server')
+server.run('go install myproject/server')
```
nicks: this would be pretty slow in practice, because it has to download all the dependencies on install. it seems like a weird example to use. A better example might be a python or nodejs or ruby one with a run trigger

docs/nested_fastbuild.md
maiamcc: ...you're totally right. let me take another pass.

maiamcc: @nicks better? (I want to leave SOME context b/c i imagine the audience for this doc is "people who have been frustrated with the old syntax of fast build" -- when we decide we want to roll this out to more people, the instructions will look slightly different/have less context on the old syntax.)

PR #77, opened 2019-03-15
docs: add .tiltignore

landism:
"other features" feels like the "util" of TOCs, but:
1. All of the existing "Using Tilt" sections are explicitly about the Tiltfile.
2. There will presumably shortly be other non-tiltfile features worth documenting (like no auto build).
3. When there's enough in here to pull some out as a more cohesive page, we can do that.


landism:
> there are two changes in your compiled output

whoops, fixed in #78

PR #80, opened 2019-03-15
add json functions to API docs

jazzdan:


api/api.py
```diff
@@ -205,3 +205,14 @@ def k8s_kind(kind: str, api_version: str=None, *, image_json_path: Union[str, Li
   """
   pass
 
+JSONType = Union[
```
jazzdan: This is a little bit of a lie but it doesn't seem worth it to get more specific than this. And from my Googling around it might not even be possible in Python's type system.

jazzdan:
@nicks :sweat_smile: yup just did that!

PR #82, opened 2019-03-20
api: add default_registry

landism:


landism:
> it would also be good to have a standalone doc

filed ch1943

PR #90, opened 2019-03-26
docs: add bazel guide

jazzdan:
This is a very rough draft and I feel that I should edit it down. Open to suggestions!

### DIFF
--- a/docs/building_images_with_bazel.md
+++ b/docs/building_images_with_bazel.md
@@ -1,9 +1,9 @@
 ---
-tilt: Extending Tilt With No Pull Requests
+tilt: Building Images with Bazel
 layout: docs
 ---
 
-# Extending Tilt with No Pull Requests
+# Building Images with Bazel
 
 There's nothing more frustrating than using a tool that works perfectly except for _one_ small thing. Even if the tool is open source it can be a daunting task to get a change accepted to it.
 
@@ -17,9 +17,7 @@ _Note: Before diving in to this guide you should have run through [Tilt's Gettin
 Here's an [example repository](https://github.com/windmilleng/bazel_example) that contains two services (snack and vigoda) that are built with Bazel. The current development workflow is to make some changes to the code (contained in the `snack` or `vigoda` directories) then run the appopriate bazel rule to put the service in to my Kubernetes cluster. For example if I made a change to `snack/main.go` I would then run `bazel run //:snack-server.apply`. This command will build my Go code, put it in to an image, put that image in to my YAML, and run `kubectl apply` to update my Kubernetes cluster with the new YAML. That's a lot of stuff but Bazel handles it all by understanding the transitive dependencies of `snack-server`.
 
 ## Writing the Tiltfile
-So how can we take this sophisticated workflow and combine its reproducability and dependency tracking with Tilt's lightning fast updates and heads up display?
-
-Let's remember that Tiltfiles only need two things in order to provide a great experience: Kubernetes YAML and Docker Images. To get our service up in running Tilt we just need to figure out how to ask Bazel for those two things separately.
+Tiltfiles only need two things in order to provide a great experience: Kubernetes YAML and Docker Images. To get our service up in running Tilt we just need to figure out how to ask Bazel for those two things separately.
 
 ### Getting YAML
 So when we run `bazel run //:snack-server.apply` what does Bazel actually do? To find out let's look at the `:snack-server` rule definition.
@@ -87,7 +85,7 @@ Easy enough, now in the main body of our Tiltfile we can call it like so:
 k8s_yaml(bazel_k8s(":snack-server"))
 ```
 
-And if we Tilt up we should see that Tilt calls Bazel to build the YAML, and by extension its image and then Tilt takes that YAML and deploys it to the cluster. We get all the awesomeness of the HUD with three lines of code in a Tiltfile.
+And if we Tilt up we should see that Tilt calls Bazel to build the YAML, and by extension its image and then Tilt takes that YAML and deploys it to the cluster.
 
 ### Building Images
 Now for the second component: images. Let's follow the same strategy and figure out the Bazel command that will just build an image and then integrate it in to the Tiltfile.
@@ -112,14 +110,27 @@ We know what the image will be ahead of time because of Bazel's deterministic na
 The third argument, the empty list, is a list of dependencies. We'll fill that in later. For now let's wrap this in a function called `bazel_build` and add it to the main section of our Tiltfile
 
 ```python
+def bazel_build(image, target):
+  custom_build(
+    image,
+    'bazel run ' + target,
+    [],
+    tag="image",
+  )
 k8s_yaml(bazel_k8s(":snack-server"))
-bazel_image('bazel/snack', '//snack:image')
+bazel_build('bazel/snack', '//snack:image')
 ```
 
 Now Tilt can get YAML from Bazel, get an image from Bazel, insert it into the YAML and track its status in the Kubernetes cluster.
 
 ## Dealing with Dependencies
-There's a problem with our Tiltfile as it stands: it only picks up changes if the Tiltfile changes. It doesn't understand what dependencies go in to generating the YAML _or_ the code dependencies that go in to generating the docker image. Luckily for us Bazel makes it easy to ask for those dependencies and Tilt makes it easy to tell it about those dependencies. Let's start by getting the dependencies for the YAML and then move on to the dependencies for the docker image.
+If you started Tilt now, it would appear to work! Run `tilt up` and all the images get deployed to your cluster.
+
+Then you make a change to main.go. Nothing gets updates. Why not?
+
+It's not enough for Tilt to know how to build & deploy an image. It needs to know what files should trigger a new build and deploy.
+
+Luckily for us Bazel makes it easy to ask for those files and Tilt makes it easy to tell it about those files.
 
 ### YAML Dependenices
 [`bazel query`](https://docs.bazel.build/versions/master/query-how-to.html) is the functionality that we'll use to ask Bazel about the various dependencies of certain targets. Before we do that there are two types of dependencies that Tilt cares about: dependencies for the Tiltfile and dependencies for images.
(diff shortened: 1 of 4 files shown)


docs/extending_tilt_with_bazel_support.md
```diff
@@ -0,0 +1,289 @@
+---
+tilt: Extending Tilt With No Pull Requests
+layout: docs
+---
+
+# Extending Tilt with No Pull Requests
+
+There's nothing more frustrating than using a tool that works perfectly except for _one_ small thing. Even if the tool is open source it can be a daunting task to get a change accepted to it.
```
nicks: i feel like there's so much throat-clearing here that it's really hard for a reader to answer the question "Does this doc apply to me?" That's the main question a user skimming this is trying to answer.

maybe something like

```
By default, Tilt builds images with Docker and deploys with Kubectl

But what if you don't even have a Dockerfile to give Tilt? What if you build and deploy your images with Bazel?

It's a lot simpler than you'd think. Let's see why.
```

docs/extending_tilt_with_bazel_support.md
```diff
@@ -0,0 +1,289 @@
...
+
+This tells Tilt to run the command (`bazel run //snack:image`) and to assert after it has completed that an image named bazel/snack will exist with the tag "image". If it doesn't then the build fails.
+
+We know what the image will be ahead of time because of Bazel's deterministic nature. Bazel uses a formula for determining the name of the image that it will push: "bazel/" + the directory that the build file is in". This image is tagged with the name of the rule that created it, which in our case is "image".
+
+The third argument, the empty list, is a list of dependencies. We'll fill that in later. For now let's wrap this in a function called `bazel_build` and add it to the main section of our Tiltfile
+
+```python
+k8s_yaml(bazel_k8s(":snack-server"))
+bazel_image('bazel/snack', '//snack:image')
```
nicks: where does the bazel_image function come from?

docs/extending_tilt_with_bazel_support.md
```diff
@@ -0,0 +1,289 @@
...
+  [],
+  tag="image",
+)
+```
+
+This tells Tilt to run the command (`bazel run //snack:image`) and to assert after it has completed that an image named bazel/snack will exist with the tag "image". If it doesn't then the build fails.
+
+We know what the image will be ahead of time because of Bazel's deterministic nature. Bazel uses a formula for determining the name of the image that it will push: "bazel/" + the directory that the build file is in". This image is tagged with the name of the rule that created it, which in our case is "image".
+
+The third argument, the empty list, is a list of dependencies. We'll fill that in later. For now let's wrap this in a function called `bazel_build` and add it to the main section of our Tiltfile
```
nicks: did you intend to include the bazel_build function here?

docs/extending_tilt_with_bazel_support.md
```diff
@@ -0,0 +1,289 @@
...
+
+```python
+k8s_yaml(bazel_k8s(":snack-server"))
+bazel_image('bazel/snack', '//snack:image')
+```
+
+Now Tilt can get YAML from Bazel, get an image from Bazel, insert it into the YAML and track its status in the Kubernetes cluster.
+
+## Dealing with Dependencies
+There's a problem with our Tiltfile as it stands: it only picks up changes if the Tiltfile changes. It doesn't understand what dependencies go in to generating the YAML _or_ the code dependencies that go in to generating the docker image. Luckily for us Bazel makes it easy to ask for those dependencies and Tilt makes it easy to tell it about those dependencies. Let's start by getting the dependencies for the YAML and then move on to the dependencies for the docker image.
```
nicks: this paragraph is hard to follow. you need to already be immersed in how Tilt works to understand it.

Maybe better to tell the story from the user's perspective?

Like,

```
If you started Tilt now, it would appear to work! Run `tilt up` and all the images get deployed to your cluster.

Then you make a change to main.go. Nothing gets updates. Why not?

It's not enough for Tilt to know how to build & deploy an image. It needs to know what files should trigger a new build and deploy.

Luckily for us Bazel makes it easy to ask for those files and Tilt makes it easy to tell it about those files.
```

docs/extending_tilt_with_bazel_support.md
```diff
@@ -0,0 +1,289 @@
...
+
+### YAML Dependenices
```
nicks: i might even axe the word 'dependencies' completely from this doc. these aren't abstract dependencies. They're just files to the user.

docs/extending_tilt_with_bazel_support.md
jazzdan: whoops, this is supposed to be `bazel_k8s`.

docs/extending_tilt_with_bazel_support.md
jazzdan: Ooh yes I like this.

docs/extending_tilt_with_bazel_support.md
jazzdan: While that's true, a Bazel user might be more familiar with the nomenclature?

docs/building_images_with_bazel.md
```diff
@@ -0,0 +1,300 @@
...
+      l = l[2:]
+
+    path = l.replace(":", "/")
+    if path.startswith("/"):
+      path = path[1:]
+
+    watch_file(path)
+    watched_files.append(path)
+
+  return watched_files
```
landism: Does nothing use this? Would it be clearer if we dropped the return value?

docs/building_images_with_bazel.md
```diff
@@ -0,0 +1,300 @@
...
+  watch_labels(source_deps)
+
+  return local("bazel run %s" % target)
+
+def bazel_build(image, target):
+  build_deps = str(local(BAZEL_BUILDFILES_CMD % target)).splitlines()
+  watch_labels(build_deps)
+
+  source_deps = str(local(BAZEL_SOURCES_CMD % target)).splitlines()
+  source_deps_files =bazel_labels_to_files(source_deps)
```
landism:
```suggestion
  source_deps_files = bazel_labels_to_files(source_deps)
```

docs/building_images_with_bazel.md
```diff
@@ -0,0 +1,300 @@
...
+  return local("bazel run %s" % target)
+```
+
+Easy enough, now in the main body of our Tiltfile we can call it like so:
+
+```python
+k8s_yaml(bazel_k8s(":snack-server"))
+```
+
+And if we Tilt up we should see that Tilt calls Bazel to build the YAML, and by extension its image and then Tilt takes that YAML and deploys it to the cluster.
```
landism:
```suggestion
And if we `tilt up` we should see that Tilt calls Bazel to build the YAML, and by extension its image and then Tilt takes that YAML and deploys it to the cluster.
```

docs/building_images_with_bazel.md
```diff
@@ -0,0 +1,300 @@
...
+
+### Building Images
+Now for the second component: images. Let's follow the same strategy and figure out the Bazel command that will just build an image and then integrate it in to the Tiltfile.
+
+From reading about rules_k8s I've learned that it uses rules_docker under the hood for building images. Another [quick glance at rules_docker's documentation](https://github.com/bazelbuild/rules_docker#using-with-docker-locally) reveals that `bazel run //:targetname` will load the image in to the configured docker daemon. Perfect, that's all we need!
```
landism:
```suggestion
From reading about rules_k8s, we learn that it uses rules_docker under the hood for building images. Another [quick glance at rules_docker's documentation](https://github.com/bazelbuild/rules_docker#using-with-docker-locally) reveals that `bazel run //:targetname` will load the image in to the configured docker daemon. Perfect, that's all we need!
```

docs/building_images_with_bazel.md
```diff
@@ -0,0 +1,300 @@
+---
+tilt: Building Images with Bazel
+layout: docs
+---
+
+# Building Images with Bazel
+
+There's nothing more frustrating than using a tool that works perfectly except for _one_ small thing. Even if the tool is open source it can be a daunting task to get a change accepted to it.
+
+We recognize that everyone's set up is different and that we cannot foresee all possible features and use cases that folks will need. Software is complicated! As a result we designed Tilt to be customizable from the beginning. If there's functionality you wish Tilt had you can often add it yourself, right in your Tiltfile, no external pull requests necessary!
```
landism:
```suggestion
We recognize that everyone's set up is different and that we cannot foresee all possible features and use cases that folks will need. Software is complicated! As a result we designed Tilt to be customizable from the beginning. If there's functionality you wish Tilt had you can often add it yourself, right in your Tiltfile, no code changes or pull requests necessary!
```

This might be nitty, but the "external" in "external pull requests" rubbed me a bit weirdly - it's framed relative to tilt devs, who are not the intended audience.

docs/building_images_with_bazel.md
```diff
@@ -0,0 +1,300 @@
...
+
+Let's explore this with a simple feature request: I want Tilt to support the [Bazel](https://bazel.build/) build system. I can hear you now: but that's not simple at all! It's actually a lot simpler than you'd think, let's see why.
+
+_Note: Before diving in to this guide you should have run through [Tilt's Getting Started Guide](tutorial.html) and the deep dive in to [Tiltfile concepts](tiltfile_concepts.html). A familiarity with the Bazel build system will also clear a lot of things up but isn't necessary. Ready? Let's get started!_
```
landism: the "A familiarity with the Bazel build system" feels a bit weird if this doc is now "how to use tilt with your existing bazel build" rather than "here's an example of implementing a thing in the Tiltfile!"

docs/integrating_bazel_with_tilt.md
```diff
@@ -0,0 +1,203 @@
...
+```python
+k8s_yaml(bazel_k8s(":snack-server"))
+```
+
+And if we `tilt up` we should see that Tilt calls Bazel to build the YAML and then Tilt takes that YAML and deploys it to the cluster. If the image exists in the registry Kubernetes should schedule the resource
+
+### Building Images
+Let's follow the same strategy and figure out the Bazel command that will just build an image and then integrate that in to the Tiltfile.
+
+rules_k8s uses rules_docker under the hood for building images. Another [quick glance at rules_docker's documentation](https://github.com/bazelbuild/rules_docker#using-with-docker-locally) reveals that `bazel run //:targetname` will load the image in to the configured Docker daemon. That's all we need thanks to `custom_build`. Custom build takes an image name and a command that should produce that image. After the command is executed Tilt asserts that it exists, and retags it so it can track it.
```
nicks: put rules_k8s and rules_docker in code quotes?

PR #91, opened 2019-03-28
api: add docs for live_update

landism:
This is just api docs to ensure we're on the same page. I've filed ch2047 for friendlier non-api docs.

This won't be merged until after the implementation is in master.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -7,7 +7,60 @@ class Blob:
 
    To wrap a string as a blob, call ``blob(my_str)``"""
 
-def docker_build(ref: str, context: str, build_args: Dict[str, str] = {}, dockerfile: str = "Dockerfile", dockerfile_contents: Union[str, Blob] = "") -> None:
+class LiveUpdateStep:
+  """A step in the process of performing a LiveUpdate on an image's container.
+
+  See the LiveUpdate documentation, coming soon!
+  """
+  pass
+
+def fall_back_on(files: Union[str, List[str]]) -> LiveUpdateStep:
+  """Specify that any changes to the given files will cause Tilt to *fall back* to a
+  full image build (rather than performing a live update).
+
+  ``fall_back_on`` step(s) may only go at the beginning of your list of steps.
+
+  (Files must be a subset of the files that we're already watching for this image;
+  that is, if any files fall outside of DockerBuild.context or CustomBuild.deps,
+  an error will be raised.)
+
+  Args:
+      files: a string or list of strings of files. If relative, will be evaluated relative to the Tiltfile. Tilt compares these to the local paths of edited files when determining whether to fall back to a full image build.
+  """
+  pass
+
+def sync(local_path: str, remote_path: str) -> LiveUpdateStep:
+  """Specify that any changes to `localPath` should be synced to `remotePath`
+
+  May not follow any `run` steps in a `live_update`.
+
+  Args:
+      localPath: A path relative to the Tiltfile's directory. Changes to files matching this path will be synced to `remotePath`.
+          Can be a file (in which case just that file will be synced) or directory (in which case any files recursively under that directory will be synced).
+      remotePath: container path to which changes will be synced. Must be absolute.
+  """
+  pass
+
+def run(cmd: str, trigger: Union[List[str], str] = []) -> LiveUpdateStep:
+  """Specify that the given `cmd` should be executed when updating an image's container
+
+  May not precede any `sync` steps in a `live_update`.
+
+  Args:
+    cmd: A shell command.
+    trigger: If the ``trigger`` argument is specified, the build step is only run on changes to the given file(s). Paths relative to Tiltfile.
+  """
+  pass
+
+def container_restart() -> LiveUpdateStep:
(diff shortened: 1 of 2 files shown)


api/api.py
```diff
@@ -294,3 +294,64 @@ def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable
...
   pass
+
+class LiveUpdateStep:
+    """A step in the process of performing a LiveUpdate on an image's container.
+
+    See :meth:`live_update`.
+    """
+    pass
+
+def workdir(dir: str) -> LiveUpdateStep:
```
maiamcc: also sync "dest"'s, if relative, are evaluated relative to this. maybe worth a "see also" to docker's WORKDIR command?

maiamcc: specify that we only allow one and it must come first

api/api.py
```diff
@@ -294,3 +294,64 @@ def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable
...
+
+    - 0 or 1 :meth:`workdir`
+    - 0 or more :meth:`sync`
+    - 0 or more :meth:`run`
+    - 0 or 1 :meth:`container_restart`
+
+    When a file changes:
+
+    1. If it matches any of the paths in `full_rebuild_trigger`, a full rebuild + deploy will be executed.
+    2. Otherwise, if it matches any of the local paths in `sync` steps, a live update will be executed.
```
maiamcc: ...a live update will be executed:
- copy any changed sync'd files to the container
- execute relevant run's
- if container_restart provided, restart the container

(or something)

api/api.py
maiamcc: oh hmm you list this below... idk, maybe worth listing here as well? (and similarly under container_restart?) 🤷‍♀️

api/api.py
```diff
@@ -294,3 +294,65 @@ def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable
...
+    pass
+
+def run(cmd: str, trigger: Union[List[str], str] = []) -> LiveUpdateStep:
+    """Specify that the given `cmd` should be executed when updating an image's container
+
+    May not precede any `sync` steps in a `live_update`.
+
+    Args:
+      cmd: A shell command.
+      trigger: If the ``trigger`` argument is specified, the build step is only run on changes to the given file(s).
```
jazzdan:
```suggestion
      trigger: If the ``trigger`` argument is specified, the build step is only run on changes to the given file(s). Paths relative to Tiltfile.
```

api/api.py
```diff
@@ -294,3 +294,65 @@ def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable
...
+
+    """
+    pass
+
+def live_update(image: str, steps: List[LiveUpdateStep], full_rebuild_trigger: Union[str, List[str]] = []) -> None:
+    """Specify how to update containers running `image`.
+
+    The list of steps must be, in order:
+
+    - 0 or more :meth:`sync`
```
nicks: meth? :eyes:

landism: this makes `sync` a hyperlink to the method `sync`, instead of just monospaced text

https://stackoverflow.com/a/22714510

nicks: :breaking-bad-this-is-not-meth-dot-gif:

maiamcc:
I don't need a full re-review, but but appreciate a quick skim in re: layout and organization

maiamcc:
cc @landism i can't request you as a reviewer of your own PR but hiii

api/api.py
```diff
@@ -7,7 +7,60 @@ class Blob:
 
    To wrap a string as a blob, call ``blob(my_str)``"""
 
-def docker_build(ref: str, context: str, build_args: Dict[str, str] = {}, dockerfile: str = "Dockerfile", dockerfile_contents: Union[str, Blob] = "") -> None:
+class LiveUpdateStep:
+  """A step in the process of performing a LiveUpdate on an image's container.
```
landism: Is it worth linking all the LiveUpdateStep constructors from here, or let the docs handle that?

api/api.py
```diff
@@ -7,7 +7,60 @@ class Blob:
...
+  pass
+
+def run(cmd: str, trigger: Union[List[str], str] = []) -> LiveUpdateStep:
+  """Specify that the given `cmd` should be executed when updating an image's container
+
+  May not precede any `sync` steps in a `live_update`.
+
+  Args:
+    cmd: A shell command.
+    trigger: If the ``trigger`` argument is specified, the build step is only run on changes to the given file(s). Paths relative to Tiltfile.
```
landism:
```suggestion
    trigger: If the ``trigger`` argument is specified, the build step is only run when there are changes to the given file(s). Paths relative to Tiltfile.
```

(my first read of the old wording was like "oh it'll somehow intelligently only run go build on the files that changed - wait how can it do that?")

landism:
LGTM

api/api.py
maiamcc: UHHHH :-/ i'm still holding out hope that we can get the nesting to work. let me try that once we get the stuff merged, if not i'll go back and link stuff

PR #92, opened 2019-04-02
docs: change optimization doc from fast_build to live_update

landism:
This will be landed in sync with https://github.com/windmilleng/tiltdemo/pull/5

### DIFF
--- a/docs/live_update.md
+++ b/docs/live_update.md
@@ -39,9 +39,9 @@ live_update(dm1_img_name,
   ])
 ```
 
-This looks similar to the `Tiltfile` in previous tutorials, but in addition to building
-with `docker_build()`, it contains `live_update()`. Let's zoom
-in on that part of the configuration.
+This looks similar to the `Tiltfile` in previous tutorials, but in addition to specifying
+how to build the image with `docker_build()`, it specifies how to update the running
+image with `live_update()`. Let's zoom in on that part of the configuration.
 
 
 ```python
@@ -66,14 +66,18 @@ The `sync` method copies a file or directory from outside your container to insi
 In this case, we copy the directory `./cmd/demoserver1` (relative to the Tiltfile) into
 the container filesystem.
 
-While Tilt is running, it watches all files in `./cmd/demoserver1`. If they change, it copies the file
-into the container.
+The normal `docker_build` behavior is to watch all files in the docker build context (`.`),
+and any time one changes, to do an image build and redeploy. The `sync` here says that, if
+the changed file matches `cmd/demoserver1`, to instead do a `live_update` - Tilt will copy
+the changed files into the container and execute any appropriate `run` or `restart_container`
+steps, without actually building or pushing a docker image or performing a k8s deploy.
 
 * `run('go install github.com/windmilleng/tiltdemo/cmd/demoserver1')`
 
 The `run` method runs shell commands inside your container.
 
-Every time a file changes, Tilt will run this command again.
+Every time a `live_update` runs (i.e., when a file matching a `sync` changes), Tilt will run
+this command again.
 
 One of the major build optimizations that Tilt does is to keep the container around, and
 start the command inside the running container.
--- a/docs/live_update.md
+++ b/docs/live_update.md
@@ -79,13 +79,12 @@ The `run` method runs shell commands inside your container.
 Every time a `live_update` runs (i.e., when a file matching a `sync` changes), Tilt will run
 this command again.
 
-One of the major build optimizations that Tilt does is to keep the container around, and
-start the command inside the running container.
-
-This is much closer to how we normally run commands for local development. Real humans
-don't delete all their code and re-clone it from git every time we need to do a new build!
-Instead, we re-run the command in the same directory. Modern tools then take advantage of local caches;
-Tilt runs commands with the same approach, but inside a container.
+One of the major build optimizations here is that Tilt runs your command inside the existing
+container, instead of building from scratch to create a fresh image. This is much closer to how
+we normally run commands for local development. Real humans don't delete all their code and
+re-clone it from git every time we need to do a new build! Instead, we re-run the command in
+the same directory. Modern tools then take advantage of local caches; Tilt runs commands with
+the same approach, but inside a container.
 
 * `restart_container()`
 
@@ -93,8 +92,8 @@ This specifies that the container should be restarted after the other update ste
 applied. Any changed files stay around and, in the case of k8s, the pod stays where it is. This is
 effectively just re-starting the service in the existing container.
 
-For languages/frameworks with hot reloading (i.e., they can pick up code changes without
-restarting the process), like node or flask, this step is unnecessary.
+For languages/frameworks like Node or Flask that have hot reloading (i.e., they can pick up code
+changes without restarting the process), this step is unnecessary.
 
 In this guide, we explored just a few of the functions we can use in a `Tiltfile`
 to keep your build fast. For even more functions and tricks,
--- a/docs/live_update.md
+++ b/docs/live_update.md
@@ -84,7 +84,7 @@ start the command inside the running container.
 
 This is much closer to how we normally run commands for local development. Real humans
 don't delete all their code and re-clone it from git every time we need to do a new build!
-We re-run the command in the same directory. Modern tools then take advantage of local caches.
+Instead, we re-run the command in the same directory. Modern tools then take advantage of local caches;
 Tilt runs commands with the same approach, but inside a container.
 
 * `restart_container()`
--- a/docs/live_update.md
+++ b/docs/live_update.md
@@ -86,7 +86,7 @@ Tilt runs commands with the same approach, but inside a container.
 * `restart_container()`
 
 This specifies that the container should be restarted after the other update steps have been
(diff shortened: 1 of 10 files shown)


docs/live_update.md
```diff
@@ -30,53 +30,43 @@ The `Tiltfile` at the root of the repo contains this example:
...
+    run('go install github.com/windmilleng/tiltdemo/cmd/demoserver1'),
+    restart_container(),
+  ])
 ```
 
-This looks similar to the `Tiltfile` in previous tutorials, but instead of building
-with `docker_build()`, it contains `fast_build()`. Let's zoom
-in on that part of the function.
+This looks similar to the `Tiltfile` in previous tutorials, but in addition to building
+with `docker_build()`, it contains `live_update()`. Let's zoom
```
maiamcc: either "in addition to building with... it updates with..." or "in addition to calling... it also includes a call to..."

i.e. talk about either the build behavior OR the tiltfile, not both at once.

docs/live_update.md
```diff
@@ -30,53 +30,43 @@ The `Tiltfile` at the root of the repo contains this example:
...
+The `sync` method copies a file or directory from outside your container to inside of your container.
 
-The `add` method copies a directory from outside your container to inside of your container.
-
-In this case, we copy the directory `./cmd/demoserver` (relative to the Tiltfile) into
+In this case, we copy the directory `./cmd/demoserver1` (relative to the Tiltfile) into
 the container filesystem.
 
-While Tilt is running, it watches all files in `./cmd/demoserver`. If they change, it copies the file
+While Tilt is running, it watches all files in `./cmd/demoserver1`. If they change, it copies the file
```
maiamcc: not quite -- it watches all the files in `.`, but if they match `./cmd/demoserver1`, it runs the liveupdate steps, otherwise does a full build. put in some more here?

docs/live_update.md
```diff
@@ -85,14 +75,23 @@ The `run` method runs shell commands inside your container.
...
 
 This is much closer to how we normally run commands for local development. Real humans
 don't delete all their code and re-clone it from git every time we need to do a new build!
 We re-run the command in the same directory. Modern tools then take advantage of local caches.
 Tilt runs commands with the same approach, but inside a container.
 
+* `restart_container()`
+
+This specifies that the container should be restarted after the other update steps have been
+applied. The files stay around and, in the case of k8s, the pod stays where it is. This is
```
maiamcc:
```suggestion
applied. Any changed files stay around and, in the case of k8s, the pod stays where it is. This is
```

docs/live_update.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Optimizing a Tiltfile
```
maiamcc: thought for future: if we want `LiveUpdate` to in face be a "simple case", maybe we should move this content into the main tutorial?

docs/live_update.md
```diff
@@ -0,0 +1,101 @@
...
+The `sync` method copies a file or directory from outside your container to inside of your container.
+
+In this case, we copy the directory `./cmd/demoserver1` (relative to the Tiltfile) into
+the container filesystem.
+
+The normal `docker_build` behavior is to watch all files in the docker build context (`.`),
+and any time one changes, to do an image build and redeploy. The `sync` here says that, if
+the changed file matches `cmd/demoserver1`, to instead do a `live_update` - Tilt will copy
+the changed files into the container and execute any appropriate `run` or `restart_container`
+steps, without actually building or pushing a docker image or performing a k8s deploy.
```
maiamcc: 🙌 perf.

docs/live_update.md
```diff
@@ -0,0 +1,101 @@
...
+
+Every time a `live_update` runs (i.e., when a file matching a `sync` changes), Tilt will run
+this command again.
+
+One of the major build optimizations that Tilt does is to keep the container around, and
+start the command inside the running container.
+
+This is much closer to how we normally run commands for local development. Real humans
+don't delete all their code and re-clone it from git every time we need to do a new build!
+We re-run the command in the same directory. Modern tools then take advantage of local caches.
```
maiamcc:
```suggestion
Instead, we re-run the command in the same directory. Modern tools then take advantage of local caches;
```

docs/live_update.md
```diff
@@ -0,0 +1,101 @@
...
+steps, without actually building or pushing a docker image or performing a k8s deploy.
+
+* `run('go install github.com/windmilleng/tiltdemo/cmd/demoserver1')`
+
+The `run` method runs shell commands inside your container.
+
+Every time a `live_update` runs (i.e., when a file matching a `sync` changes), Tilt will run
+this command again.
+
+One of the major build optimizations that Tilt does is to keep the container around, and
```
maiamcc: this paragraph is weird. Maybe:
"One of the major optimizations here is that Tilt runs your command _in the existing container_ -- so you get the benefit of any caches, existing artifacts, etc."

docs/live_update.md
```diff
@@ -0,0 +1,101 @@
...
+We re-run the command in the same directory. Modern tools then take advantage of local caches.
+Tilt runs commands with the same approach, but inside a container.
+
+* `restart_container()`
+
+This specifies that the container should be restarted after the other update steps have been
+applied. Any changed files stay around and, in the case of k8s, the pod stays where it is. This is
+effectively just re-starting the service in the existing container.
+
+For languages/frameworks with hot reloading (i.e., they can pick up code changes without
```
maiamcc: Hmm reads weird? maybe: 

"For languages/frameworks like Node or Flask with hot reloading (i.e., that can pick up code changes without
restarting the process), this step is unnecessary."

or:

"For languages/frameworks with hot reloading (i.e., that can pick up code changes without
restarting the process -- e.g. Node or Flask), this step is unnecessary."

maiamcc:
Updated tutorial to reflect the new syntax. Also, made a separate LiveUpdate reference (to put the notes on live-update logic that we don't have room for in the API docs, now that LiveUpdate isn't its own method)

i'm open to suggestions of where else to put this info -- i just like the idea of having a page to link to that's technical nitty gritty as opposed to a tutorial. wdyt? cc @landism @nicks

maiamcc:
(when I merge #91 i'll go back and put links to the constructor func api references in `live_update_reference.md`)

docs/live_update_tutorial.md
```diff
@@ -0,0 +1,100 @@
...
+
+In the [Tutorial](tutorial.html), we introduced the `docker_build()` function.
+This function builds a Docker image. Tilt will watch the inputs to the
+image, and rebuild it every time they change.
+
+This works well for interpreted languages like JavaScript and Python
+where you can add the files and go. For servers that need to be compiled,
+it would be too slow to recompile from scratch every time.
+
+That's why Tilt has a function `live_update()` for lightning-fast local
```
landism: not a function

docs/live_update_tutorial.md
```diff
@@ -0,0 +1,100 @@
+---
+title: Optimizing a Tiltfile
+layout: docs
+---
+This tutorial looks at a `Tiltfile` with build optimizations.
+We explain what they do, and why you would want to use them.
+
+##### (This is a tutorial that walks you through a sample project. If you're looking for technical specs and details, check out the [LiveUpdate Reference](live_update_reference.html) docs.)
```
landism: "LiveUpdate Reference" kinda feels like a typo. Would "Live Update Reference" or "live_update refeference" work?

docs/live_update_reference.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Live Update Reference
+layout: docs
+---
+##### (This doc provides the technical specifications of Tilt's `LiveUpdate` functionality. For a tutorial that walks you through a sample project, see [Optimizing a Tiltfile](live_update_tutorial.html).)
+
+When specifying how to build an image (via `docker_build()` or `custom_build()`), you may optionally pass the `live_update` arg.
+
+`live_update` takes a list of `LiveUpdateSteps` that tell Tilt how to update a running container in place. (This allows us to skip Docker build/push, take advantages of on-disk caches, and generally give you a faster development experience.)
```
landism: A couple of things that might be useful and/or repetitive here:
1. "update a running container in place instead of paying the cost of building a new image and redeploying"
2. a "See here for details" link to a blog post - or maybe we're waiting for a new post from DB that's more live_update than fast_build?

docs/live_update_reference.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Live Update Reference
+layout: docs
+---
+##### (This doc provides the technical specifications of Tilt's `LiveUpdate` functionality. For a tutorial that walks you through a sample project, see [Optimizing a Tiltfile](live_update_tutorial.html).)
```
landism: if we're calling this a technical spec, I'd be inclined to:
1. make at least the first instance of each method a hyperlink to the api doc
2. mention the run trigger behavior somewhere (define below's use of "relevant")

docs/live_update_reference.md
maiamcc: ah yep, waiting on a blog post

docs/live_update_tutorial.md
maiamcc: meh, sure

PR #103, opened 2019-04-11
api: document workload_to_resource_function

landism:


api/api.py
```diff
@@ -401,3 +401,40 @@ def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable
     live_update: set of steps for updating a running container (see `Live Update documentation <live_update_reference.html>`_).
   """
   pass
+
+
+class k8sObjectID:
```
jazzdan: Our other classes begin with a capital letter FWIW.

landism: good call!

PR #113, opened 2019-04-16
jekyll/Tiltfile: api.py can benefit from LiveUpdate (hacky)

maiamcc:
as the note in the Tiltfile says, this solution is hacky. A change to
`api.py` results in TWO in-place builds -- global logs look like this:

```
[Tiltfile] Beginning Tiltfile execution
[Tiltfile] Running `"make api"`
[Tiltfile] Successfully loaded Tiltfile
1 changed: [api/api.py]
──┤ Rebuilding: docs-site ├────────────────────────────────────────────
  → Updating container…
Copying 1 file(s) to container: 170b4fff0d
- '/Users/maia/code/go/src/github.com/windmilleng/tilt.build/api/api.py' --> '/tmp/api.py'
  → Container updated!
1 changed: [docs/_includes/api.html]
──┤ Rebuilding: docs-site ├────────────────────────────────────────────
  → Updating container…
Copying 1 file(s) to container: 170b4fff0d
- '/Users/maia/code/go/src/github.com/windmilleng/tilt.build/docs/_includes/api.html' --> '/docs/_includes/api.html'
  → Container updated!
docs-site   ┊       Regenerating: 1 file(s) changed at 2019-04-16 19:57:12
docs-site   ┊                     _includes/api.html
docs-site   ┊        Jekyll Feed: Generating feed for posts
docs-site   ┊                     ...done in 1.7629268 seconds.
```

a combination of [local build steps](https://app.clubhouse.io/windmill/epic/2056/local-build-steps-spec) and ordered LiveUpdate step execution could
save us from this... but this is still faster than the old world, where a change to
`api.py` resulted in a full rebuild, so we'll keep this for now even tho it's ugly.


Tiltfile
```diff
@@ -18,6 +20,13 @@ docker_build('tilt-site', 'src', dockerfile='site.dockerfile',
 
 docker_build('docs-site', '.', dockerfile='docs.dockerfile',
              live_update = [
+               # we don't care about this being on the container but we don't want it to trigger a full rebuild,
+               # just toss it somewhere so it's accounted for by a LiveUpdate step
+               # This is hacky. Right now a change to api.py kicks off TWO container updates, one
+               # to sync api.py (which we don't care about), and one when the local('make api') call
+               # above changes docs/_includes/api.html -- but it's still faster than a full rebuild ¯\_(ツ)_/¯
+               sync('./api/api.py', '/tmp/api.py'),
```
nicks: I feel like this comment points to a bigger problem with the api - full-rebuild-on is kind of unhelpful most of the time, the more common case is an ignore param

Have you considered putting api.py into tiltignore?

maiamcc: yes I tried that -- the result is that changes to api.py don't kick off tiltfile re-execution :-/

PR #116, opened 2019-04-17
add an faq item for microk8s

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/microk8s:

4728cde18989bb24bf37fc2e5a8cd271fdec6602 (2019-04-17 15:16:57 -0400)
add an faq item for microk8s

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -173,3 +173,32 @@ Tilt reads the same `KUBECONFIG` environment variable that `kubectl` uses. This 
 sets the path to a Kubernetes config.
 
 <script src="/assets/js/links.js" async></script>
+
+### Q: How do I set up Tilt to use MicroK8s?
+
+If you're using Linux, [microk8s](https://microk8s.io/)
+is a fast, low-overhead way to run Kubernetes locally for development.
+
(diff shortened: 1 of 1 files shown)


nicks:
We're getting pretty close to the point where it might make sense to have a "choose-your-own-adventure" of different cluster configurations, but at least now we'll have something to point people at

docs/faq.md
```diff
@@ -173,3 +173,32 @@ Tilt reads the same `KUBECONFIG` environment variable that `kubectl` uses. This
...
+To install it, run
+
+```
+$ sudo snap install microk8s --channel=1.13/stable --classic
+```
+
+To add microk8s to the list of clusters you can deploy to with `kubectl`, run
+
+```
+$ microk8s.kubectl config view --raw > $HOME/.kube/config
```
jazzdan: Does this always work? My understanding is that sometimes the microk8s config changes when you restart it?

nicks: Resolved offline. It might be worth doing a blog post on all the ways to set kubecon fig context

PR #130, opened 2019-05-08
docs: add upgrade instructions

landism:
This is basically just the installation doc, plus a link to the releases page, minus a bunch of stuff that isn't relevant to upgrades.

This will make bumping the version for docs more annoying, since now there are two pages to update. I'm tempted to write a script.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -1,0 +1,41 @@
+---
+title: Upgrading
+layout: docs
+---
+
+You can find a list of Tilt releases on [Tilt's GitHub Releases page](https://github.com/windmilleng/tilt/releases).
+
+On MacOS
+--------
+
+### Option A) If you installed Tilt via Homebrew
+
+```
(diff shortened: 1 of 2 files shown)


docs/upgrade.md
```diff
@@ -0,0 +1,41 @@
+---
+title: Upgrading
+layout: docs
+---
+
+You can find a list of Tilt releases on [Tilt's GitHub Releases page](https://github.com/windmilleng/tilt/releases).
```
nicks: should we also put a link to this doc in the sidebar? or in the install guide?

landism: I kinda figured this page wasn't interesting enough to link to from the sidebar, and it was really only interesting as a landing page, but:

1. maybe it is actually useful?
2. in general, it's annoying when I land on a page and navigate away from it and try to navigate back, but it's not in the menu.

added

PR #131, opened 2019-05-10
docs: update node example readme

drubin:
The api now uses an `[]` array param but the docs still list multiple params

Small doc fix.


jazzdan:
Thanks @drubin!

PR #155, opened 2019-05-28
api: note that fast_build is deprecated rather than linking to non-existant docs

### LINKED ISSUE #153: Broken link to fast_build
https://docs.tilt.dev/api.html#api.fast_build

links to

https://docs.tilt.dev/fast_build.html

which 404s

jazzdan:
Fixes #153


maiamcc:
Good catch!

PR #158, opened 2019-06-04
api: document ignore/only parameter to docker_build

jazzdan:


api/api.py
```diff
@@ -70,10 +70,12 @@ def docker_build(ref: str, context: str, build_args: Dict[str, str] = {}, docker
...
-    dockerfile: path to the Dockerfile to build
-    dockerfile_contents: raw contents of the Dockerfile to use for this build
+    build_args: build-time variables that are accessed like regular environment variables in the ``RUN`` instruction of the Dockerfile. See `the Docker Build Arg documentation <https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg>`_.
+    dockerfile: path to the Dockerfile to build.
+    dockerfile_contents: raw contents of the Dockerfile to use for this build.
     live_update: set of steps for updating a running container (see `Live Update documentation <live_update_reference.html>`_).
+    ignore: set of file patterns that will be ignored. Ignored files will not trigger builds and will not be included in images. Follows the `dockerignore syntax <https://docs.docker.com/engine/reference/builder/#dockerignore-file>`_.
+    only: set of file patterns that will cause a build to be triggered. All other changes will be ignored. Inverse of ignore parameter. Follows the `dockerignore syntax <https://docs.docker.com/engine/reference/builder/#dockerignore-file>`_. Note that ignores take precedence over this parameter.
```
dbentley: Might be good to have a caveat about how the implementation is complex and in Docker, so if you ahve very tricky cases you should reach out to us.

PR #171, opened 2019-06-13
jekyll/tiltfile: changes to api.py shouldn't kick off a full rebuild; all we need is a live update to sync the files changed when we ran 'make api'

maiamcc:
Improvements to tiltfile to make it easier to iterate on the API Reference.

(I originally did something like this in #113 but failed to merge it into
master properly whooops -- and anyway, this solution is much cleaner!
hooray for `docker_build.ignore`!)


Tiltfile
```diff
@@ -16,15 +18,15 @@ docker_build('tilt-site', 'src', dockerfile='site.dockerfile',
                run('bundle install', trigger=['src/Gemfile', 'src/Gemfile.lock'])
              ])
 
-docker_build('docs-site', '.', dockerfile='docs.dockerfile',
+docker_build('docs-site', '.', dockerfile='docs.dockerfile', ignore=["./api/api.py"],
```
nicks: should this be in the other docker_build as well?

maiamcc: nah, it doesn't fall within the build context of the other dockerfile.

maiamcc: Ha actually, after merging master, yes it should!

PR #181, opened 2019-06-14
docs: tweak update control

landism:
1. reword a paragraph:
  1. the "now" made it feel time-sensitive, and more appropriate for an announcement than a doc
  2. maybe silly, but "but it's up to you" kinda made it feel more like a burden than an empowerment
2. add a screenshot and some text explaining what the user should expect in the UI. The screenshot might need to be updated after upcoming UI tweaks, but c'est la vie.


docs/manual_update_control.md
```diff
@@ -39,3 +39,11 @@ k8s_resource('foo')
...
 k8s_resource('bar', trigger_mode=TRIGGER_MODE_AUTO)
 ```
+
+Now, when you run Tilt, the web UI's sidebar will look something like this:
+
+<div class="block u-margin1_5">
+ <img src="assets/img/update-control-button-in-menu.png">
+</div>
+
+When you make changes to "foo", instead of them being automatically applied, Tilt will simply indicate unapplied changes by the asterisk to the right of `foo` in the sidebar, and it will not automatically apply those changes, instead waiting until you click the apply button to the left of `foo`.
```
jazzdan: This sentence feels a little run-on.

PR #206, opened 2019-06-25
Update "Don't Touch Prod" example

redeemed2011:
The documented example for "Don't Touch Prod" fails to function in my environment. However, whenever I add `()` to align with the API reference for `k8s_context()`, the example code functions as intended.


nicks:
thanks for the fix!

redeemed2011:
It was a massive undertaking.

I am very much enjoying the tool. It is so very much better than anything else I have come up with or seen or used. :) Thank you, Tilt, team also for supporting docker-compose.

PR #216, opened 2019-07-12
api: document custom_build.ignore param

maiamcc:
Hello @mariabelenvivanco, @jazzdan,

Please review the following commits I made in branch maiamcc/custombuild.ignore:

734aa5c7af30bc44256912d0a43aac0928ae7fd7 (2019-07-12 12:05:13 -0400)
api: document custom_build.ignore param


api/api.py
```diff
@@ -435,7 +435,7 @@ def add_fast_build() -> FastBuild:
     """Returns a FastBuild that is associated with the image that was built from a ``custom_build``. When the container needs to be rebuilt it will be built using the ``CustomBuild``. Otherwise update will be done with the ``FastBuild`` instructions. """
     pass
 
-def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable_push: bool = False, live_update: List[LiveUpdateStep]=[], entrypoint: str="") -> CustomBuild:
+def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable_push: bool = False, live_update: List[LiveUpdateStep]=[], ignore: Union[str, List[str]] = [], entrypoint: str="") -> CustomBuild:
```
mariabelenvivanco: I think the variable name is "ignores" instead of "ignore"

mariabelenvivanco: I think the variable is name is "ignores" rather than "ignore"

maiamcc: hmm, [doesn't look like it](https://github.com/windmilleng/tilt/blob/71d688dd63b727a1a29fa0b0ad4e33cebf24f658/internal/tiltfile/docker.go#L242)?

mariabelenvivanco: Oo ok I see- it looks like we call it "ignores" internally after we parse the args, my bad! Otherwise LGTM!

PR #275, opened 2019-09-11
Fix container restart method name

drubin:
I struggled a bit to figure out why my `container_restart` method wasn't working it turns out the API docs don't seem to match the code. 

https://docs.tilt.dev/api.html#api.restart_container . 

Hopefully this fixes that.


nicks:
thanks! lgtm

PR #289, opened 2019-09-20
api: add local_resource

jazzdan:


api/api.py
```diff
@@ -597,3 +597,19 @@ def enable_feature(feature_name: str) -> None:
   Args:
     feature_name: name of the feature to enable
   """
+  pass
+
+def local_resource(name: str, cmd: str, deps: Union[str, List[str]] = None, trigger_mode: TriggerMode = TRIGGER_MODE_AUTO) -> None:
+  """Configures `cmd` to run on the _host_ machine, not in a remote cluster.
+
+  If `deps` is set then `cmd` is only run if one of the files specified changes.
```
maiamcc:
```suggestion
  If `deps` is set then `cmd` is run whenever one of the files specified changes.
```
(consider something like this because technically that initial run of this command is not triggered by a change to one of these files)

api/api.py
```diff
@@ -597,3 +597,19 @@ def enable_feature(feature_name: str) -> None:
...
+
+  If `trigger_mode` is set then the local resource obeys the trigger mode.
```
maiamcc: this seems kinda tautological, imo can cut this and just explain in the `args` section below

api/api.py
```diff
@@ -597,3 +597,19 @@ def enable_feature(feature_name: str) -> None:
...
+
+  Args:
+    name: will be used as the new name for this resource
+    cmd: command to be executed on host machine
+    deps: a list of files or directories to be added as dependencies to this cmd. Tilt will watch those files and will run the cmd when they change.
+    trigger_mode: may be one of ``TRIGGER_MODE_AUTO`` or ``TRIGGER_MODE_MANUAL``
```
maiamcc: link to docs like other func w/ `trigger_mode` do?
```suggestion
    trigger_mode: one of ``TRIGGER_MODE_AUTO`` or ``TRIGGER_MODE_MANUAL``. For more info, see the
      `Manual Update Control docs <manual_update_control.html>`_.
```

PR #298, opened 2019-09-30
docs: document disable_snapshots

landism:


api/api.py
```diff
@@ -622,3 +622,14 @@ def local_resource(name: str, cmd: str, deps: Union[str, List[str]] = None, trig
       `Manual Update Control docs <manual_update_control.html>`_.
   """
   pass
+
+def disable_snapshots() -> None:
+    """Disables Tilt's "snapshots" feature, hiding it from the UI.
```
jazzdan: Should we link to the (soon to be revamped) snapshot doc here?

PR #302, opened 2019-10-02
docs: add FAQ for snapshots privacy

landism:


docs/snapshots.md
```diff
@@ -43,6 +43,13 @@ You can view and delete all of the snapshots associated with your account on Til
 ### Q: What data do you store for each snapshot?
 For each snapshot we store the entirety of your Tilt's state, including all logs and build history. In addition we store two pieces of metadata: the time that the snapshot was taken and the user, if any, that created it.
 
+### Q: How private are snapshots?
+Snapshots are accessible to anyone who has the link, which includes an unguessable ID.
+If this makes you or your team nervous, you can disable them for your project with the
+[`disable_snapshots`](https://docs.tilt.dev/api.html#api.disable_snapshots)
+directive in your Tiltfile. If you're interested in using snapshots, but want
+access control, please [let us know](https://tilt.dev/contact)!
```
dbentley: I think phrasing it as "makes you nervous" can, itself, make you nervous. It could be better to separate the "is it private?" (not yet; reach out to us!) from the "i can't use SaaS" (disable it)

Maybe another question like:

### Q: My company doesn't allow using SaaS; how can I disable Snapshots for my team?

PR #310, opened 2019-10-17
api: Add documentation for os.environ

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/env:

52b199545d14651347045ff64a9d26ff0c365fb3 (2019-10-17 18:10:47 -0400)
api: Add documentation for os.environ


nicks:
looks a little bit awkward right now because we have one module with one property but hopefully that will change!

src/_data/docs.yml
```diff
@@ -48,6 +48,8 @@
   items:
   - title: Tiltfile API Reference
     href: api.html
+  - title: Tiltfile API - os module
```
maiamcc: we're probably going to want to do something different with modules in the sidebar, rather than throwing all these links in with other links -- maybe their own subsection, or another level of nesting? but this is fine for now.

nicks: ya, totally agree. if we have enough of these, we're going to want to change the flow here.

PR #325, opened 2019-10-31
flags proposal

dbentley:


docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+* `tilt up a b d`: run A, B, and D but not C
+* `tilt flags a b`: change running Tilt to run A and B but neither C nor D.
+* `tilt up` (a second time): run all services
+
+#### Tiltfile
+```python
+flags.define_string_list("to-run", args=True, sticky=False)
+cfg = flags.parse()
+
+flags.exclude_all_resources()
```
landism: Is there a reason the `resource` builtins are on `flags`, or is it just because we're adding them at the same time? What if I want to use them instead with `read_json` or `local`?

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+for r in cfg['to-run']:
+  flags.include_resource(r)
```
landism: Unfortunately, `resource_deps` has complicated this - does this include `r` and its transitive deps, or just `r`?
The former is probably more generally desirable, while the latter leaves an escape hatch.
Maybe a bool with a default value?

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+You want to be able to write a doc that tells users to run `tilt up consumer` the first time and know that's enough that they'll be set up. Running Tilt subsequent times without arguments should default to the mode they already have set.
+
+#### Command-Lines
+* `tilt up`: run all services
+* `tilt up consumer`: run consumer services and record mode
+* `tilt up` (a second time): run consumer services
+
+#### Tiltfile
+```python
+flags.define_string_list("to-run", args=True)
```
landism: Is this analogous to an existing arg parsing library? would it make sense to start with something that could expand to, e.g., [argparse syntax](https://docs.python.org/3/library/argparse.html)? It wouldn't be great to invent our own syntax from scratch, either in terms of wasting time figuring things out ourselves, or in terms of potential lost preexisting user familiarity.

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+if 'b' in cfg['to-edit']:
+  docker_build('b', './b')
+if 'c' in cfg['to-edit']:
+  docker_build('c', './c')
+if 'd' in cfg['to-edit']:
+  docker_build('d', './d')
+```
+
+## How Flags Work
+Tilt stores flags in a file next to the Tiltfile called `tilt_flags.json`. When the Tiltfile executes `flags.parse()`, it loads that file as JSON. If it's the first execution after a Tilt up, the execution will merge the flags passed on the command-line with the values from the file (using the definition of flags as a guide) and write the new values to the file if necessary.
```
landism: Merging seems pretty confusing. Could we punt it and see if we meet user needs without it?

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+### More Kinds of Flags
+You should be able to define more kinds of flags. For example, a list of ints or floats, or a singly-valued string. By moving error-checking to the built-in library, you can make your Tiltfile shorter and more correct.
+
+### Better Flag Manipulation
+Your users should be able to understand what reasonable values are. This could be from help output, tab-completion, or in the Web UI.
+
+Your users should be able to modify flags in the Web UI, without having to know there's a file underneath.
+
+### Escape Hatch for Flag Parsing
+You should be able to use a custom syntax for your flags. The flags module could take a function that would parse a command-line into a JSON value.
```
landism: FWIW, I think we don't even need this much - we just need to provide a `flags()` function that returns a raw array of strings, and then the user can do whatever they want to convert that to `cfg`.

Or maybe merging complicates that, I'm still confused by how merging works.

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+if 'd' in cfg['to-edit']:
+  docker_build('d', './d')
+```
+
+## How Flags Work
+Tilt stores flags in a file next to the Tiltfile called `tilt_flags.json`. When the Tiltfile executes `flags.parse()`, it loads that file as JSON. If it's the first execution after a Tilt up, the execution will merge the flags passed on the command-line with the values from the file (using the definition of flags as a guide) and write the new values to the file if necessary.
+
+Any change to the flags file will cause Tilt to reexecute the Tiltfile. You can modify the flags file with an editor or write scripts that write values into it, and Tilt will update.
+
+A call to `tilt flags` will connect to the running Tilt instance on the server port and cause Tilt to reexecute the Tiltfile, using the passed command-line args as if it's the first execution after a Tilt up.
```
landism: At the risk of getting nitty, for both this and the line above - I might say "any change to flags will cause Tilt to reexecute the Tiltfile", and then the other mechanisms might change flags.

Which is to say, maybe some operation touches the flags file but doesn't change its contents, or I run `tilt flags` to see the current values but don't change anything, and that shouldn't necessarily run the Tiltfile.

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Tiltfile
+```python
+flags.define_string_list("to-run", args=True)
+flags.define_string_list("to-edit")
+cfg = flags.parse()
+
+# omitting the code from above Tiltfile that includes correct resources
+
+# only configure the build for services we want to edit
```
landism: I haven't put too much thought into the wording, but it might be worth explicitly calling out something like "Thanks to Tilt's existing behavior, images that don't have a `docker_build` directive will just deploy with whatever is configured in their yaml, so it's really just this easy"

docs/flags.md
landism: We've (or maybe just I've?) kind of had this notion that the order of Tiltfile builtins doesn't matter.

Structuring this as "exclude" and then multiple "includes" makes it more of an imperative thing.

Other options:
* rename exclude to something like "explicitly_include_resources()", so it sounds more like it's setting a flag that resources are not included by default
** haven't come up with a name for this that I'm perfectly happy with
* get rid of exclude entirely and make any call to include imply exclude
** feels slightly magical
* get rid of exclude entirely, change include_resource to, like, set_resources() and instead of appending to the list, constructs the list
** has much clearer semantics, but requires the user to do an extra step of assembly a list of strings before calling, but that seems pretty minor?

Or maybe I'm the only one that had that principle in my head in the first place.

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+      flags.include_resource(r)
+  else:
+    flags.include_resource(r)
+```
+
+By making the flag sticky (the default), Tilt will record the flag between runs.
+
+
+### Specify services to edit
+You have many services that your users frequently need to run, but don't expect to edit. Your YAML includes images from your CI that are suitably recently (and match what's in production or staging).
```
landism:
```suggestion
You have many services that your users frequently need to run, but don't expect to edit. Your YAML includes images from your CI that are suitably recent (and match what's in production or staging).
```

landism: or even instead of "suitably recent (and match what's in production or staging)." just "good enough" (with maybe "so you don't want to waste time building them")

docs/flags.md
dbentley: Just r. If you want more complicated behavior, you have to code it. Perhaps we have a way to get the deps of a resource?

docs/flags.md
dbentley: Because it has to do with customization. I wasn't sure whether to call the module flags or config or args or custom. Open to better ideas.

dbentley: I like set_resources()!

docs/flags.md
dbentley: In terms of how to define things in the Tiltfile or in terms of what's accepted on the command-line?

docs/flags.md
dbentley: Yeah, I like that. Will incorporate.

docs/flags.md
dbentley: Yep; will incorporate.

docs/flags.md
dbentley: We've had multiple explicit requests to store a context, and allow other tools to write the config.

docs/flags.md
dbentley: All of our other watching is based on timing, not on contents.

I agree that tilt flags --help shouldn't reinvoke.

docs/flags.md
dbentley: Yeah, let's discuss a bit more?

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+
+A call to `tilt flags` will connect to the running Tilt instance on the server port and cause Tilt to reexecute the Tiltfile, using the passed command-line args as if it's the first execution after a Tilt up.
+
+## Future Directions
+This section describes places we expect flags to go. Let us know if any of these would be particularly helpful to you and your team.
+
+### More Kinds of Flags
+You should be able to define more kinds of flags. For example, a list of ints or floats, or a singly-valued string. By moving error-checking to the built-in library, you can make your Tiltfile shorter and more correct.
+
+### Better Flag Manipulation
```
maiamcc:
```suggestion
### Better Flag Manipulation/Discovery
```
(Or make "discovery" its own section?)

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+## Examples
+
+### Run only some services
+Your app has many services, call them A, B, C, and D. You want users to be able to run only some of them.
+
+and you want your users to be able to run only a subset of them. E.g., `tilt up` will run all services, but running `tilt up a b d` will run only the resources a, b, and d, ignoring c. Each run of Tilt will start from scratch.
+#### Command-Lines
+* `tilt up`: run all services
+* `tilt up a b d`: run A, B, and D but not C
+* `tilt flags a b`: change running Tilt to run A and B but neither C nor D.
```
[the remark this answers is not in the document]
dbentley: tilt flags --port 10350 or tilt flags --port 10351

docs/flags.md
```diff
@@ -0,0 +1,130 @@
...
+flags.exclude_all_resources()
+for arg in cfg['args']:
+  if arg in groups:
+    for r in groups[arg]:
+      flags.include_resource(r)
+  else:
+    flags.include_resource(r)
+```
+
+By making the flag sticky (the default), Tilt will record the flag between runs.
```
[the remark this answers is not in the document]
dbentley: Honestly, I think stickiness is the first thing to cut. Yes, I can talk more about this.

docs/flags.md
landism: > yeah Matt, pretty sure in this proposal you could say cfg = read_json('config.json') -- is that what you're asking about?

right, what I'm saying is that it's weird to turn around and have to pass that to something in the `flags` namespace when I'm not using flags at all

> Because it has to do with customization. I wasn't sure whether to call the module flags or config or args or custom. Open to better ideas.

`config` seems fine, or even just putting it in the global namespace

docs/flags.md
landism: I guess I don't have strong feelings on this and maybe it's more of a question for the docs PR when we get there.

docs/flags.md
landism: the former, which might affect the latter

docs/flags.md
landism: > Think I mentioned this above, but this seems like a risky assumption (i.e. that there's only one running Tilt instance)

Risky, but probably fine for now? Like, presumably a vanishingly small fraction of our users have multiple tilts, and also a relatively small fraction of our users will be using this, and we can solve that when people complain?

at some point we might want `screen(1)`-like behavior for tilt instances
i.e., screen is able to tell how many screens are running - if there's exactly one, it uses that. if there are multiple, it lists them all and tells you how to specify which one you want

landism: > All of our other watching is based on timing, not on contents.

I don't think this decision matters much either way, but fwiw I was mostly thinking of the analogue to manifest configs - if I touch a yaml file, the Tiltfile will reexecute, re-load the yaml contents into the manifest definition, observe that the manifest definition has not changed, and will not trigger a resource build, which to a user, kind of ends up feeling like watching based on contents, depending on how disruptive that project's Tiltfile execution is.

PR #341, opened 2019-11-12
local resource docs

maiamcc:
Hello @landism, @dbentley,

Please review the following commits I made in branch maiamcc/local-resource:

551455ced340e6a494731e696b142e9bf6df89d3 (2019-11-12 12:46:32 -0500)
api docs

5beeb188528ebcfeecca095c30fffad2ce08dd6c (2019-11-12 12:28:09 -0500)
docs: local resource doc


docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
+---
+title: Run Local and/or Occasional Workflows with Local Resource
+layout: docs
+---
+A Local Resource works like any other resource in your sidebar (which are most
+often some combination of Docker Build instructions and K8s Yaml)--encompasses
```
landism: nit: "K8s Yaml" capitalization feels weird, esp. given "k8s yaml" a few lines down. I think it should be "YAML" or "yaml". I don't have an opinion and don't know if we have a docs standard on k8s vs K8s (it looks like we currently have 5 of the former and 2 of the latter), but it'd be good to be consistent within this doc

landism:
```suggestion
often some combination of Docker Build instructions and K8s Yaml)--it encompasses
```

docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
...
+a unit of work, and execute in response to file changes, or
```
landism:
```suggestion
a unit of work, and executes automatically in response to file changes, or
```

docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
...
+[manually](https://docs.tilt.dev/manual_update_control.html). For your resource
+`MyGreatService`, when one of its file dependencies changes, the work it executes
```
landism:
```suggestion
`MyGreatService`, when one of its file dependencies changes, its work
```

that sounded pretty roundabout to me

docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
...
+is to build a Docker image and deploy some k8s yaml; for a Local Resource, it's
+to execute an arbitrary command on your local filesystem.
+
+# Run commands locally on file change
+
+Tilt is designed to make it easy to build images, and to sync files/run commands
```
landism: This sentence kind of makes it sound like Local Resource is some fringe feature that tilt is not designed to make easy.

Maybe something like "Most uses of Tilt focus on easily and quickly seeing your code running in the cloud."?

docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
...
+in the cloud. For some kinds of development, however, you might want to harness
+Tilt's responsiveness to file changes, but run commands locally instead. These
```
landism: "These cases often fall into the general categories of" makes this list sound more exhaustive than I think is intended. Maybe "Some common use cases we've seen this address"?

docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
...
+(because you want the resulting `yarn.lock` file to exist on your local computer
+so that you can commit it to git)
+
+```python
+local_resource('yarn', cmd='yarn install', deps=['package.json'])
+```
+- *tooling that you have locally but don't want to put on your container images*.
+For instance: rather than pulling/pushing big container images with the Go compiler,
+you'd rather compile your binary locally, and then use Live Update to quickly sync
+the new binary up to a running container in your cluster
```
landism: it's weird to have sample code for the previous one but not this one

docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
...
+# define your hello-world resource: k8s yaml + a Docker build
+k8s_yaml('hello-world.yaml')
+docker_build('hello-world', '.',
+    ignore='helloworld.proto',  # don't step on Local Resource's toes
+    live_update=[ sync('.', '/app') ]
+)
+```
+
+Note that the `docker_build` call specifies `ignore='helloworld.proto'`. This is
+because we DON'T want an edit to that proto file to kick off a Docker build.
```
landism:
```suggestion
because we DON'T want an edit to that proto file to directly kick off a hello-world build.
```

1. we do want it to kick off a build, transitively
2. the "docker build" vs "live update" bit is a bit ambiguous (is a live update a "docker build"? if we say yes, then we definitely want the "directly" if we say no, then the `ignore` is irrelevant anyway because changing the proto would only ever kick off a live update, not a docker build - also saying no is weird when it is kicking off the live update of a docker build. this is probably just more unexplored ambiguous terminology territory)

docs/local_resource.md
```diff
@@ -0,0 +1,100 @@
...
+    trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False
+)
+```
+
+(`auto_init=False` is currently only compatible with `TRIGGER_MODE_MANUAL`. If
+you'd like a Local Resource that runs automatically in response to file changes
+but does NOT run on `tilt up`, [let us know](https://tilt.dev/contact).)
+
+Note that you can mix and match manual and automatic runs as you like. For instance,
+you might have a `seed-db` Local Resource. Usually, you run it manually whenever
```
landism: this seems like a weird solution to this use case, since seed_db.sh would run on startup

docs/local_resource.md
maiamcc: you are 100% correct ty

docs/local_resource.md
maiamcc: urghhh. but yeah you're right

docs/local_resource.md
maiamcc: better?

docs/local_resource.md
```diff
@@ -0,0 +1,106 @@
+---
+title: Run Local and/or Occasional Workflows with Local Resource
+layout: docs
+---
+Each entry in your Tilt sidebar is a **resource**---a unit of work managed by Tilt. (Until
```
landism: "Until this point" / "have been" make sense in a blog with a publish date, but less sense in a doc

maiamcc: fair nough

PR #352, opened 2019-11-16
homebrew: no need to brew tap, brew windmilleng/tap/tilt is enough

maelvls:
Just a small improvement: users of Homebrew do not need to care about the `brew tap` step since `brew install windmilleng/tap/tilt` will implicitely run `brew tap` if the tap has not already been cloned.

Just want to say: awesome project! 🤩

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -90,7 +90,6 @@ The installer first checks if you can install Tilt with Homebrew. If you'd prefe
 to run Homebrew manually, run:
 
 ```bash
-brew tap windmilleng/tap
 brew install windmilleng/tap/tilt
 ```
 


nicks:
thanks! 

do you know if this behavior was recently changed in homebrew? All the documentation i found implied that you needed to explicitly run `brew tap` but it was also kind of hand-wavey

cc @landism , can you confirm this works? I don't have my mac today.

maelvls:
Not very easy to find (https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap#installing):

> ### Installing
> If it’s on GitHub, users can install any of your formulae with `brew install user/repo/formula`. Homebrew will automatically add your `github.com/user/homebrew-repo` tap before installing the formula. `user/repo/formula` points to the `github.com/user/homebrew-repo/**/formula.rb` file here.
>
> If they want to get your tap without installing any formula at the same time, users can add it with the `brew tap user/repo` command.

landism:
WFM!

PR #357, opened 2019-11-26
api/docs: update tiltfile concepts with new understanding of resource, clarify k8s_resource api docs

maiamcc:
Matt Hodgekins at Datadog got tripped up b/c he didn't realize Tilt
automatically did most of the k8s resource config for you, asked us
to clarify our docs around this. While I was in here, I updated
Tiltfile Concepts: Resources to be more in line with our new definition
of "resource" (i.e. one that includes both K8s resources and local resources)

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -223,11 +223,28 @@ def trigger_mode(trigger_mode: TriggerMode):
 TRIGGER_MODE_AUTO = type('_sentinel', (TriggerMode,),
                  {'__repr__': lambda self: 'TRIGGER_MODE_AUTO'})()
 
+def dc_resource(name: str, trigger_mode: TriggerMode = TRIGGER_MODE_AUTO, resource_deps: List[str] = []) -> None:
+  """Configures the Docker Compose resource of the given name. Note: Tilt does an amount of resource configuration
+  for you(for more info, see `Tiltfile Concepts: Resources <tiltfile_concepts.html#resources>`_); you only need
+  to invoke this function if you want to configure your resource beyond what Tilt does automatically.
+
+  Args:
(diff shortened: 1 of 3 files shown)


api/api.py
```diff
@@ -227,7 +227,9 @@ def k8s_resource(workload: str, new_name: str = "",
                  port_forwards: Union[str, int, List[int]] = [],
                  extra_pod_selectors: Union[Dict[str, str], List[Dict[str, str]]] = [],
                  trigger_mode: TriggerMode = TRIGGER_MODE_AUTO, resource_deps: List[str] = []) -> None:
-  """Configures a kubernetes resources
+  """Configures the Kubernetes resource of the given name. Note: Tilt does an amount of resource configuration
```
nicks: I don't think this comment is accurate. Most people use k8s_reosurce for port-forwards, which Tilt does not do for you

nicks: I might split this into two comments: one here that just something like "Resource names are calculated automatically, as described in [link]", and then add a comment to "extra_pod_selectors" (or whatever parameter the user was confused about) with something like "You should use this parameter IF [you're in edge case X, Y, or Z]"

maiamcc: > Configures the Kubernetes resource of the given name. Tilt assembles (and names) Kubernetes resources automatically, as described in `Tiltfile Concepts: Resources <tiltfile_concepts.html#resources>`_). Calling ``k8s_resource`` is *optional*; you can use this function to configure port forwarding for your resource, to rename it, or to adjust any of the other settings specified below.

Better?

nicks: did you intend to push up an update?

maiamcc: i was gonna push after you OK'd the copy :)

PR #358, opened 2019-11-26
docs: add notes on helm subcharts

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/helmreq:

7e68f84957d56149143a49eb194e5b412df103dd (2019-11-26 12:54:16 -0500)
docs: add notes on helm subcharts


nicks:
it might also make sense for the helm() built-in to do this automatically, but i suspect we'll need this note either way for people that are using local()

PR #366, opened 2019-12-11
docs: tiltfile config

landism:


docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+for arg in cfg.get('args', []):
+  if arg in groups:
+    resources += groups[arg]
+  else:
+    # also support specifying individual services instead of groups, e.g. `tilt up a b d`
+    resources.append(arg)
+config.set_enabled_resources(resources)
+```
+
+### Specify services to edit
```
maiamcc: hmm this seems unclear w/o context but I'm struggling to come up w/ how to phrase it...
> Run many services, specify a subset to edit

> Specify services to edit, run rest from pre-built image

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+You have many services that your users frequently need to run, but don't expect
+to edit. Your YAML includes good enough images. (Perhaps they're from a suitably
+recent CI invocation, or refer to standard images.)
```
maiamcc: > Your YAML includes good enough images. (Perhaps they're from a suitably recent CI invocation, or refer to standard images.)

try:

> Your YAML references pre-existing images (perhaps they're from a suitably
recent CI invocation, or refer to standard images) such that the user can run the app without doing a Docker Build via Tilt

maiamcc: Maybe add something like:

> This pattern allows you to run the majority of your services from pre-existing images (avoiding the overhead of a Docker Build), while using Tilt to watch files and quickly update services you're actively developing on.

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+* `tilt up -- --to-edit b`: run all services, editing b (the '--' indicates the
+  end of `tilt up` options and the start of user-defined settings)
+* `tilt up -- consumer --to-edit b --to-edit c`: run consumer services and set
+  up b and c for editing.
+#### Tiltfile
+```python
+config.define_string_list("to-run", args=True)
+config.define_string_list("to-edit")
+cfg = config.parse()
+# omitting the code from above Tiltfile that includes correct resources
```
maiamcc: this isn't super clear, looks like a comment. maybe:

```
for arg in cfg.get('args', []):
    ... # omitting the code from above...
```

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+# Only configure the build for services we want to edit
+# Thanks to Tilt's existing behavior, an image that is mentioned in YAML but has no build set
+# will use the existing tag.
+to_edit = cfg.get('to-edit', [])
```
maiamcc: is it easier or harder to read if you do:
```
def docker_build_for_service(svc):
    docker_build(svc, './{}'.format(svc))

for svc in to_edit:
    docker_build_for_service(svc)

File: docs/user_config.md [hunk 5]
```diff
@@ -0,0 +1,150 @@
...
+if 'c' in to_edit:
+  docker_build('c', './c')
+if 'd' in to_edit:
+  docker_build('d', './d')
+```
+
+## The Config File
+In addition to config settings coming from command-line args, Tilt will read
+them from a `tilt_config.json` in the same directory as the `Tiltfile`, if one
+exists. It should contain a single JSON dict, where the keys are config setting
```
maiamcc:
```suggestion
exists. This file should contain a single JSON object, where the keys are config setting
```

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+names and the values are their values (e.g., `{"to-edit": ["b", "c"]}`).
+As with any other file read by the Tiltfile, if this config file is changed
+while Tilt is running, Tilt will pick up that change and reexecute the Tiltfile.
+If the same setting is specified in both the config file and the command-line
+args, the value from the args takes precedent.
```
maiamcc:
```suggestion
args, the value from the args takes precedence.
```

maiamcc: but also I don't know if we want to encourage using the config file for this sort of ephemeral/frequently changed arg?

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Changing args at runtime
+While Tilt is running, if the user decides they need a different set of
```
maiamcc: why isn't this in 2nd person like the rest of the doc?

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+user args (i.e., args handled by config.parse, not args to tilt itself like
+`--hud`) than what they passed to `tilt up`, they can use `tilt flags` to
+replace them. `tilt flags --clear` tells Tilt to clear the set of user args in
+use.
+Note that settings from user args always take precedent over settings from the
```
maiamcc:
```suggestion
Note that settings from user args always take precedence over settings from the
```

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+Your users should be able to understand what reasonable values are. This could
+be from help output, tab-completion, or in the Web UI.
+Tilt should provide a way to manipulate tilt_config.json.
+Your users should be able to modify the config in the Web UI.
+### Escape Hatch for Argument Parsing
+You should be able to use a custom syntax for command-line arguments. The config
+module could take a function that would parse a command-line into JSON.
+### Interaction with load/include
+Tiltfiles that are used via `load`/`include` should be able to define and use
+config settigs.
```
maiamcc: (mention the current behavior)

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
...
+* `tilt up`: run all services
+* `tilt up a b d`: run A, B, and D but not C
+* `tilt flags a b`: change running Tilt to run A and B but neither C nor D.
+* `tilt up`: run all services
+
+#### Tiltfile
+```python
+config.define_string_list("to-run", args=True)
+cfg = config.parse()
+config.set_enabled_resources(cfg.get('to-run', []))
```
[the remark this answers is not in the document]
landism: > or is it just another way to achieve the default behavior?

at the end of the description of this example: "This is the same as the default `tilt up` behavior."

I've added it a couple more times to hopefully make it clear.

> does tilt flags a b work if you're just using the default behavior?

yes

docs/user_config.md
landism: Hm, the 2nd person for the rest of the doc is the Tiltfile author, while the user here is anyone using that Tiltfile. The doc also has several passages along the lines of "You want to do X for your users".

This is a post-hoc explanation so I don't know if it's just a rationalization, but it seems reasonable to me?

docs/user_config.md
landism: you missed the typo!!!!

landism: Hmm, do you think that's important? I'd figured the obvious inference was "currently they cannot define and use config settings", but maybe that leaves open "well can they define config settings but not use them? use config settings but not define them?"

landism: (I ask because I think describing the current behavior would really overwhelm this bullet with cognitive load - you're not allowed to define new settings in any file after any file has called config.parse, and you're not allowed to call config.parse from multiple Tiltfiles in different directories. It seems more to the point just to give people the sense that they shouldn't go there)

docs/user_config.md
maiamcc: aah. sure, that's fine

docs/user_config.md
maiamcc: eh, fair point -- that's fine then, leave as is

landism:
note: proposed some new naming in https://github.com/windmilleng/tilt/pull/2671 and will update this PR once that's merged

docs/user_config.md
```diff
@@ -0,0 +1,150 @@
+---
+title: User Config
```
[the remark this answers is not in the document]
landism: changed to "tiltfile config"

maiamcc:
@landism does this want another review or do you feel good?

landism:
> @landism does this want another review or do you feel good?

all good!

PR #367, opened 2019-12-12
docs: document tiltfile config module api

landism:
half of the changes here are to get module names to show up in the api docs

For os.environ, we show "os" in the TOC but not in the detailed description:
image

This is fine since that's the only member of the "os" module and there are no other entries in the "Data" section, but it was confusing for "config" since you'd scroll through the list of methods and there's just a "parse" with no indication that it actually needs to be invoked as "config.parse".

The links to the config docs point at #366


deploy/api.dockerfile
```diff
@@ -4,29 +4,29 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
...
+RUN pip3 install sphinx==1.8.3 sphinx_rtd_theme==0.4.2 sphinx-autodoc-typehints==1.6.0 recommonmark==0.5.0 beautifulsoup4
 
 RUN mkdir -p /src
 WORKDIR /src
 
 ADD api .
 RUN make html
 
-RUN cat /src/_build/html/*.html | hgrep "dl.function" > functions.html
+RUN ./remove_prefix.py /src/_build/html/*.html | hgrep "dl.function" > functions.html
```
nicks: i wonder if we should move this all to a script

PR #375, opened 2020-01-02
docs: add note on disable BuildKit in FAQ

jazzdan:
Fixes windmilleng/tilt#2710

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -109,7 +109,7 @@ if it even tries to pull an image from a remote server.
 [BuildKit](https://github.com/moby/buildkit) is a new build engine in
 Docker for building container images.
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -110,7 +110,7 @@ if it even tries to pull an image from a remote server.
 Docker for building container images.
 
 Tilt will automatically enable Buildkit if your local Docker installation
-supports it.
+supports it and if the `DOCKER_BUILDKIT` environment variable is **not** set to `0`.
```
nicks: I might phrase this as "if you want to disable buildkit manually, set DOCKER_BUILDKIT=0"

to avoid the negative

PR #383, opened 2020-01-13
blog: the image field should now be the full url. This makes the SEO plugin work correctly

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/headerimage:

868a197a6ceefb95cb9e062170722071fa3240f6 (2020-01-13 18:57:59 -0500)
blog: the image field should now be the full url. This makes the SEO plugin work correctly


nicks:
actually, going to merge this so i can see if it works correctly on the live site

PR #386, opened 2020-01-16
Options documentation proposal

dbentley:
I'd like to get consensus on this among the x10d team and send it to prospective users.


docs/flags.md
```diff
@@ -0,0 +1,206 @@
+---
+title: Tiltfile Flags and Args
+layout: docs
+---
+
+
+Tools offer flags and args to support the wide range of users needs and preferences, and dev workflows are no different. Devs on your team will want options in how they use Tilt; some are long-standing personal preferences and others will change in the middle of working on a Pull Request. With Tilt's user configuration system, you can improve a Tiltfile that encodes one standard setup to power the breadth of workflows your team needs:
```
jazzdan: This sentence, and this doc, feels weird in that it focuses on Laura in a way that our other docs don't. It might be unnecessarily exclusionary to readers.

docs/flags.md
```diff
@@ -0,0 +1,206 @@
...
+* Maintainers choose the options to expose that make sense for their project
+* Users can easily set and change options, without having to learn a Tiltfile
+* Tiltfile extensions expose the choices so they can power arbitrary customizations
```
jazzdan: What are Tiltfile extensions?

docs/flags.md
```diff
@@ -0,0 +1,206 @@
...
+
+This doc describes how you, the Dev Experience engineer who owns the Tiltfile, can offer this to your user, the App Developer who runs Tilt while they code. In order, we'll cover:
```
jazzdan: Is this a doc or is it a guide?

docs/flags.md
```diff
@@ -0,0 +1,206 @@
...
+for arg in cfg['args']:
+  if arg in groups:
+    resources += groups[arg]
+  else:
+    resources.append(r)
+config.set_resources(resources)
+```
+
+### Specify services to edit
+You have many services that your users frequently need to run, but don't expect to edit. Your YAML includes good enough images. (Perhaps they're from a suitably recent CI invocation, or refer to standard images.)
```
landism:
```suggestion
You have many services that your users frequently need to run, but don't expect to edit. You specify good-enough images in the YAML so your users can just pull them instead of building them. (Perhaps they're from a suitably recent CI invocation, or refer to standard images.) You add a `--to-edit` arg that allows users to specify which images should be built locally.
```

docs/flags.md
```diff
@@ -0,0 +1,206 @@
...
+```python
+config.define_string_list("args", args=True)
+cfg = config.parse()
+if cfg['args']:
+  config.set_resources(cfg['args'])
+```
+
+## Future Directions
+This section describes places we expect flags to go. Let us know if any of these would be particularly helpful to you and your team.
+
```
landism: do you want to pitch stickiness here?

docs/flags.md
dbentley: I agree it's geared towards that user, and intentionally. I think exclusionary has a negative connotation. Do you think someone else would not read it? I figured they'd read it understanding the perspective. I'm not sure how to write it well for another user.

docs/flags.md
dbentley: Yeah, I'm not sure what to do with this sentence; parallelism is hard. Let me think a bit.

docs/flags.md
dbentley: What's the difference?

jazzdan: A guide is like a walk through or a tutorial, a doc describes functionality.

docs/flags.md
```diff
@@ -0,0 +1,206 @@
...
+* Define options.
+* Get option values.
+* (Optional) Set results.
+
+### Define Options
+You define an option by calling `config.define_string_list`. (We plan to support more types beyond list of strings in the future, but are starting with just one type). This allows users to pass that as a flag on the command-line or as a field in `tilt-config.json`. You can also specify one option that is populated from any remaining args. In code:
+```
+# Flags
+# Tiltfile
+config.define_string_list('foo')
```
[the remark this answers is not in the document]
dbentley: Yup, that's a great point.

docs/flags.md
dbentley: I'm not familiar with this distinction; I think of doc as the parent type, and maybe there's a difference between guide and reference. Have we used this before?

docs/flags.md
```diff
@@ -0,0 +1,206 @@
...
+config.define_string_list("to-run", args=True)
+cfg = config.parse()
+
+config.set_resources(cfg['to-run'])
+```
+
+### Run a defined set of services
+Your app has many services, call them A, B, C, and D. You have defined modes you want users to be able to start (e.g. consumer vs. enterprise) and you don't want them to have to remember which service is necessary for which mode.
+
+You want to be able to write a doc that tells users to run `tilt up consumer` the first time and know that's enough that they'll be set up. Running Tilt subsequent times without arguments should default to the mode they already have set.
```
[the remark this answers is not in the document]
dbentley: Cruft!!

docs/flags.md
dbentley: I'd be very very content if we just had "store args in file instead of in memory", which I think is a good experience and simpler.

landism: maybe my question should be "is there something in that vein on which you'd like to solicit feedback here?"

docs/tiltfile_config.md
```diff
@@ -3,24 +3,26 @@ title: Tiltfile Config
...
-  usage.
-* Describe how this is implemented. These details will help you understand how
-  to write tooling that interacts with this system and describe its behavior
-  to teammates.
-* Future work (we see how it fits in but don't plan to implement until there's
-  concrete demand).
+Tools offer flags and args to support the wide range of users needs and preferences, and dev workflows are no different. Devs on your team will want options in how they use Tilt; some are long-standing personal preferences and others will change in the middle of working on a Pull Request. With Tilt's user configuration system, you can improve a Tiltfile that encodes one standard setup to power the breadth of workflows your team needs:
+* Maintainers choose the options to expose that make sense for their project
+* Users can easily set and change options, without having to learn a Tiltfile
+* Tiltfile extensions expose the choices so they can power arbitrary customizations
```
landism: do we want this here now?

docs/tiltfile_config.md
```diff
@@ -102,7 +104,7 @@ if 'd' in to_edit:
   docker_build('d', './d')
 ```
 
-## The Config File
+## How Config Works
```
landism: this section is actually specifically about the file and not generally about how configuration works

docs/tiltfile_config.md
```diff
@@ -3,24 +3,26 @@ title: Tiltfile Config
...
+*) Setting options as a user
+*) Examples of common configuration
+*) How Config Works
+*) Future work (we see how it fits in but don't plan to implement until there's concrete demand).
+
+## Setting Options as a User
+Because users have different needs around options, Tilt supports several ways of setting options:
+
+* Jump into a mode quickly by adding flags and args on the command: `tilt up -- --key value`.
+* Store long-lived preferences in a file named `tilt-config.json` so you don't have to remember to pass options on each invocation.
```
landism:
```suggestion
* Store long-lived preferences in a file named `tilt_config.json` so you don't have to remember to pass options on each invocation.
```

docs/tiltfile_config.md
```diff
@@ -3,24 +3,26 @@ title: Tiltfile Config
...
+* Change options without having to restart/rebuild/redeploy: `tilt args -- --key value2` updates the running Tilt and reloads the config with the new value.
+* Set options in existing scripts outside of Tilt. For example, a new-hire onboarding tool that provisions capacity and requests permissions can write that info to `tilt-config.json`.
```
landism:
```suggestion
* Set options in existing scripts outside of Tilt. For example, a new-hire onboarding tool that provisions capacity and requests permissions can write that info to `tilt_config.json`.
```

docs/tiltfile_config.md
```diff
@@ -3,24 +3,26 @@ title: Tiltfile Config
...
+This doc describes how you, the Dev Experience engineer who owns the Tiltfile, can offer this to your user, the App Developer who runs Tilt while they code. In order, we'll cover:
+*) Setting options as a user
+*) Examples of common configuration
+*) How Config Works
+*) Future work (we see how it fits in but don't plan to implement until there's concrete demand).
+
+## Setting Options as a User
+Because users have different needs around options, Tilt supports several ways of setting options:
+
+* Jump into a mode quickly by adding flags and args on the command: `tilt up -- --key value`.
```
landism: I think the most popular related use case we've had requested is resource grouping - might be worth listing here

PR #389, opened 2020-01-22
tiltfile: document quiet flag for local()

tmc:
This reflects https://github.com/windmilleng/tilt/pull/2808

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -312,8 +312,13 @@ def load(path: str, *args):
     create_namespace('frontend')
   """
 
-def local(cmd: str) -> Blob:
-  """Runs cmd, waits for it to finish, and returns its stdout as a ``Blob``."""
+def local(cmd: str, quiet: bool = False) -> Blob:
+  """Runs cmd, waits for it to finish, and returns its stdout as a ``Blob``
+
+  Args:
+    cmd: Command to run.
+    quiet: If set to True, skips printing output to log.
+  """
   pass
 
 def read_file(file_path: str, default: str = None) -> Blob:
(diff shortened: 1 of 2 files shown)


api/api.py
```diff
@@ -312,8 +312,9 @@ def load(path: str, *args):
     create_namespace('frontend')
   """
 
-def local(cmd: str) -> Blob:
-  """Runs cmd, waits for it to finish, and returns its stdout as a ``Blob``."""
+def local(cmd: str, quiet: bool = False) -> Blob:
+  """Runs cmd, waits for it to finish, and returns its stdout as a ``Blob``. If ``quiet`` is set to True, skips omitting output.
```
jazzdan: Now that this command has two arguments, should we reformat it to look like the others with an arglist? e.g.

```python
def local(cmd: str, quiet: bool = False) -> Blob:
  """Runs cmd, waits for it to finish, and returns its stdout as a ``Blob``

  Args:
    cmd: Command to run
    quiet: If set to True, skips printing output to log
  """
```

jazzdan: I also think the "skips omitting output" verbiage is confusing, so I tried rewording it in the example above to: "skips printing output to log".

tmc: Yeah I meant to write emitting. I’ll clean up the signature and the language.

tmc: Yeah I meant to write emitting. I’ll clean up the signature and the language.

tmc: Yeah I meant to write emitting. I’ll clean up the signature and the language.

tmc:
@jazzdan PTAL.

jazzdan:
Thanks @tmc!

PR #405, opened 2020-02-03
api: add decode_yaml

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/decodeyaml:

10334b458f91dbc1d723bb18a754ca950af9e81a (2020-02-03 17:21:37 -0500)
api: add decode_yaml

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -423,6 +423,10 @@ def read_json(path: str, default: str = None) -> StructuredDataType:
     default: If not `None` and the file at `path` does not exist, this value will be returned."""
   pass
 
+def decode_yaml(yaml: str) -> StructuredDataType:
+  """Deserializes a given string from YAML to Starlark. Fails if the string can't be parsed as YAML."""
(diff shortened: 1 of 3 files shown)


api/api.py
```diff
@@ -423,6 +423,10 @@ def read_json(path: str, default: str = None) -> StructuredDataType:
     default: If not `None` and the file at `path` does not exist, this value will be returned."""
   pass
 
+def decode_yaml(yaml: str) -> StructuredDataType:
```
jazzdan: This is a pre-existing problem but I'm unfamiliar with `StructuredDataType`, is that a thing a pythonista would be familiar with? Should we explain it somehow?

nicks: AFACT, the python api docgen expands this to 

Return type:	Union [ Dict [ str , Any ], List [ Any ]]
https://docs.tilt.dev/api.html#api.decode_json

PR #408, opened 2020-02-04
docs: document local_resource serve

landism:


docs/local_resource.md
```diff
@@ -57,3 +57,30 @@ For more on trigger mode, [see the docs](https://docs.tilt.dev/manual_update_con
...
 but does NOT run on `tilt up`, [let us know](https://tilt.dev/contact).
+
+## serve_cmd
+
+local_resource's `serve_cmd` argument allows a local_resource to function as a
+persistent process, so you can use it for things like running services locally
+instead of in k8s, or `tail -f`.
+
+Without `serve_cmd`, a local_resource functions as a sort of batch job - when
+Tilt runs it, it expects it to terminate. While it's running, it's "in progress",
```
maiamcc: maybe "Tilt runs the command and expects it to terminate. While the command is running, the resource is 'in progress'..."

docs/local_resource.md
```diff
@@ -57,3 +57,30 @@ For more on trigger mode, [see the docs](https://docs.tilt.dev/manual_update_con
 `auto_init=False` is currently only compatible with `TRIGGER_MODE_MANUAL`. If
 you'd like a local resource that runs automatically in response to file changes
 but does NOT run on `tilt up`, [let us know](https://tilt.dev/contact).
+
+## serve_cmd
```
maiamcc: maybe worth it to elaborate on WHY we named it this -- because we imagine it being (at least in the platonic case) a command that runs a server. even if it's not always used for that, i think it'll help users to wrap their heads around the naming and what it's for if they have an example

docs/local_resource.md
```diff
@@ -57,3 +57,30 @@ For more on trigger mode, [see the docs](https://docs.tilt.dev/manual_update_con
...
+local_resource's `serve_cmd` argument allows a local_resource to function as a
+persistent process, so you can use it for things like running services locally
+instead of in k8s, or `tail -f`.
+
+Without `serve_cmd`, a local_resource functions as a sort of batch job - when
+Tilt runs it, it expects it to terminate. While it's running, it's "in progress",
+and when it finishes, it's green or red depending on its exit code.
+
+With `serve_cmd`, when the resource updates, Tilt will first run its `cmd`, if
+it is non-empty. Then, if `cmd` succeeds, Tilt will run its `serve_cmd`. As soon
```
maiamcc:
```suggestion
it is non-empty (this is often a command that prepares the system to run `serve_cmd`, e.g. building a binary). Then, if `cmd` succeeds, Tilt will run its `serve_cmd`. As soon
```

api/api.py
```diff
@@ -586,12 +586,17 @@ def enable_feature(feature_name: str) -> None:
 def local_resource(name: str, cmd: str, deps: Union[str, List[str]] = None,
                    trigger_mode: TriggerMode = TRIGGER_MODE_AUTO,
                    resource_deps: List[str] = [], ignore: Union[str, List[str]] = [],
-                   auto_init: bool=True) -> None:
-  """Configures `cmd` to run on the *host* machine (not in a remote cluster).
+                   auto_init: bool=True, serve_cmd: str = "") -> None:
+  """Configures one or more commands to run on the *host* machine (not in a remote cluster).
 
-  If `deps` is set then `cmd` is run whenever one of the files specified changes.
+  By default, local_resources are updated on ``tilt up`` and whenever any of their ``deps`` change.
```
maiamcc: sigh, "updated" feels unclear to me here :-/

mayyybe: "local resources are update (i.e. executed) on `tilt up`..."? or is that worse? :-/

docs/local_resource.md
landism: yeah, good point

though trying to justify it just makes it sound like a bad name :(

api/api.py
landism: Thanks for persisting on this annoying issue that I despaired of tackling.

I've adjusted it and it's still not great, but I think it's much better. I'm gonna go ahead and merge, but am happy to edit / discuss further if you have more thoughts.

PR #410, opened 2020-02-04
docs: misc tiltfile config fixes

landism:
based on some slack feedback

### DIFF
--- a/api/modules/config.py
+++ b/api/modules/config.py
@@ -10,8 +10,8 @@ def define_string_list(name: str, args: bool=False, usage: str="") -> None:
 
     Args:
       name: The name of the config setting
-      args: If true, any positional args (e.g., in `tilt up -- 1 2 3`,
-            this setting would get ["1" "2" "3"])
+      args: If true, this setting takes the value of any positional args (e.g.,
+            in ``tilt up -- 1 2 3``, this setting would get ``["1" "2" "3"]``)
       usage: When arg parsing fails, what to print for this setting's description.
     """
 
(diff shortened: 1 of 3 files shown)


docs/tiltfile_config.md
```diff
@@ -38,6 +38,9 @@ and starting point.
 * `tilt up`: run all services
 
 #### Tiltfile
+
+(see the [config API reference](/api.html#modules.config.define_string_list>)
```
maiamcc:
```suggestion
(See the [config API reference](/api.html#modules.config.define_string_list>)
```

docs/tiltfile_config.md
```diff
@@ -38,6 +38,9 @@ and starting point.
...
+for details on what exactly these calls are doing)
```
maiamcc:
```suggestion
for details on what exactly these calls are doing.)
```

api/modules/config.py
```diff
@@ -10,8 +10,8 @@ def define_string_list(name: str, args: bool=False, usage: str="") -> None:
 
     Args:
       name: The name of the config setting
-      args: If true, any positional args (e.g., in `tilt up -- 1 2 3`,
-            this setting would get ["1" "2" "3"])
+      args: If true, this setting takes the value of any positional args (e.g.,
```
maiamcc: ... i still have no idea what this is trying to say. can you try again? or give an example and i'll help you wordsmith?

landism: did you, um, see the example in parentheses at the end of the line? or you want another one?

and yeah I'm not sure if there are any good words to use here and was mostly leaning on the example in parentheses to make things clear

landism: ok, I think I was confused that the example didn't suffice because I hadn't considered that someone might read this without knowing what it does when `args=False`, which was silly of me.

How's it look now?

PR #413, opened 2020-02-05
docs: Add documentation on Kind local registries

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/kind:

1b3021cd8fb05bd3254e4b627882f78d5c470d88 (2020-02-05 16:20:59 -0500)
docs: Add documentation on Kind local registries

### DIFF
--- a/docs/choosing_clusters.md
+++ b/docs/choosing_clusters.md
@@ -11,12 +11,12 @@ We're here to help you figure out the right one for you.
 Beginner Level:
 
 - [Docker for Desktop](#docker-for-desktop)
+- [Kind](#kind)
 - [Microk8s](#microk8s)
 
 Intermediate Level:
 
 - [Minikube](#minikube)
-- [KIND](#kind)
 - [K3D](#k3d)
 
 Advanced Level:
@@ -29,7 +29,7 @@ Advanced Level:
 
 ## Docker for Desktop
 
-Docker for Desktop is what we recommend most often for MacOS users.
+Docker for Desktop is the easiest to get started with if you're on MacOS.
 
 In the Docker For Mac preferences, click
 [Enable Kubernetes](https://docs.docker.com/docker-for-mac/#kubernetes)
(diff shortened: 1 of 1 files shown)


docs/choosing_clusters.md
```diff
@@ -43,11 +43,37 @@ In the Docker For Mac preferences, click
...
+- Creating a new cluster is fast (~20 seconds). Deleting a cluster is even faster. 
+- Much more robust than Docker for Mac. Uses containerd instead of docker-shim. Short-lived clusters tend to be more reliable.
+- Supports a local image registry (with our [custom setup instructions](https://github.com/windmilleng/kind-local)).
+  Pushing images is fast. No fiddling with image registry auth credentials.
+- Can run in [most CI environments](https://github.com/kind-ci/examples) (TravisCI, CircleCI, etc)
+
+### Cons
+
+- The local registry setup is still new, and changing rapidly. You need to be using Tilt v0.12.0+
+- If Tilt can't find the registry, it will use the much slower `kind load` to load images.
```
maiamcc: idk if this is a con worth expressing but: yeah cluster setup is relatively fast, but you'll probably be recreating your cluster more often than on d4m, which makes it FEEL slower sometimes

docs/choosing_clusters.md
```diff
@@ -43,11 +43,37 @@ In the Docker For Mac preferences, click
...
+time also makes it a good solution for local dev. Follow these instructions for
+setting up Kind with Tilt:
+
+[Kind Setup Instructions](https://github.com/windmilleng/kind-local)
+
+### Pros
+
+- Creating a new cluster is fast (~20 seconds). Deleting a cluster is even faster. 
+- Much more robust than Docker for Mac. Uses containerd instead of docker-shim. Short-lived clusters tend to be more reliable.
+- Supports a local image registry (with our [custom setup instructions](https://github.com/windmilleng/kind-local)).
```
maiamcc: hmm, feels weird that this script lives in a separate github repo -- how hard would it be to package this as `tilt kind-cluster`?

docs/choosing_clusters.md
nicks: hmmm...that sounds like more of an objection to a particular pattern of using Kind, than an objection to Kind itself.

I recall you suggested that during the cluster-api meeting (i.e., whether it was worthwhile to recreate the cluster each time), and they pushed back pretty hard on it.

let's listen to whether other users raise this tension, and how they frame it, and maybe that will help inform how to frame it here.

docs/choosing_clusters.md
nicks: good question! Two hypotheses:

1) The specifics of this setup are still changing frequently (across Tilt versions and across Kind versions). If we hard-code this in Tilt, that means you have a weird situation where each Tilt version only works with a specific Kind version, and you run into versioning hell (oops, we can't upgrade Tilt to v0.13 because we're still on Kind 0.7)

2) People will have their own cluster setup scripts (i.e., people will want to start up Kind with the particular Kubernetes version they use in production), and will need to know how to modify them to work with Tilt.

maiamcc: I wonder if you'd get corresponding problems if we change how we match to the local registry in Tilt and someone has an out-of-date `kind-with-registry.sh`?

Anyway I think we'll need to make this more ergonomic if we want KinD to be Tilt's recommended solution to local dev, but i'm fine doing it this way for now

docs/choosing_clusters.md
maiamcc: that's fair!

PR #417, opened 2020-02-07
Fix quit command in getting started

EvanBoyle:
Looks like the doc may be out of date: 

Screen Shot 2020-02-06 at 10 12 25 PM


jazzdan:
Thanks @EvanBoyle!

PR #422, opened 2020-02-12
docs: hidden extension preview doc and API doc for load that mentions extensions

jazzdan:
This ended up being half-way between a blog post and an article, and half way between a fully written blog post draft and an outline. As a result I'm looking for high level feedback only.


docs/extensions_preview.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Extensions Preview
+layout: docs
+---
+
+Introducing Tilt Extensions! Extensions are a simple way to bundle up Tiltfile functionality that should be shared with the world.
```
landism: This opening is very focused on the perspective of the extension author, rather than the extension user, which maybe raises the question of what's the goal of this?

Some potential answers:
1. convince people to write extensions
2. convince people to go check out extensions and make use of them
3. just generally make people optimistic about Tilt's flexibility and momentum

This is just a gut feeling, but I suspect even for (1), it'd be better motivation to write this from the extension user's perspective "look, someone wrote this once and now everyone can use and learn from it!". Which maybe all the like "yeah well that's what writing code and open source is good for" might be obvious when you give it a second's thought, but it still might be worth discussing from that angle instead to make sure it's top-of-mind?

I think it also feels closer to how I would phrase the root use case of extensions, along the lines of "as a Tiltfile author, I would like a richer set of pre-written functionality available to me"

docs/extensions_preview.md
```diff
@@ -0,0 +1,45 @@
...
+
+We think that there are lots of higher level patterns that people are implementing with Tilt that could benefit from sharing, socialization and standardization. For example many users have implemented their own ways to force all kubernetes resources in to a namespace, inject sidecars or run tests. Extensions provide a way for users to package up these ideas and share them with the community.
```
landism: maybe lower level than you're looking for, but I think there's a good opportunity here to brag about our userbase and make this more positive, along the lines of "lots of our users have implement useful Tiltfile functionality that'd be great for others to be able to use as well" (we have lots of users! and they're writing great stuff!), while "many users have implemented their own ways" feels more eeyore-y "ugh our users have all this duplicate code"

docs/extensions_preview.md
```diff
@@ -0,0 +1,45 @@
...
+
+Then you could compose them together in a function:
+
+```python
+def api_server_logs():
+  api_server_pod_name = str(local('kubectl get pods --namespace kube-system -o=jsonpath="{.items..metadata.name}" -l component=kube-apiserver')).rstrip(\n)
+  local_resource('API Server Logs', '', serve_cmd='kubectl logs -f -n %s' % api_server_pod_name )
+```
+
+That's it. You've created your first Tilt extension! Simply [open a PR](https://github.com/windmilleng/tilt-extensions/compare) against the official [Tilt Extensions Repo](https://github.com/windmilleng/tilt-extensions) to make this extension available to all Tilt users. All a user needs to do now is add a simple call to `load` to their Tiltfile that references the extension:
```
landism: ok, maybe this is making this explicitly a "write your first Tilt extension" doc

maybe we want one doc for writing and one for using? or just a TOC in this doc with sections "writing an extension", "finding/using extensions"

jazzdan: Yeah I think I'll change this to writing an extension/why would you want to write an extension and then write a separate one for using extensions in a separate PR, since I think we should have some for realsies extensions to use before writing a usage guide 😅

docs/extensions_preview.md
jazzdan: Love this feedback, thanks

docs/writing_tilt_extensions.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Writing Tilt Extensions (Preview)
+layout: docs
+---
+
+Introducing Tilt Extensions! Extensions are a simple way to bundle up Tiltfile functionality that should be shared with the world.
```
landism: mentioned on slack, but as a non-blog-post, the first sentence here should probably feel less announce-y

docs/writing_tilt_extensions.md
```diff
@@ -0,0 +1,55 @@
...
+
+Lots of Tilt users have implemented useful Tiltfile functionality that'd be great for others to be able to use as well. We've seen functions that force resources in to different namespaces, inject sidecars or even run tests. We created extensions to make it easy for Tiltfile authors to share those ideas, and for new Tiltfile authors to stand on the shoulders of giants by building on top of them.
```
landism: this paragraph feels kind of announce-y, too, but I think it's fine?

docs/writing_tilt_extensions.md
```diff
@@ -0,0 +1,55 @@
...
+
+```python
+def jest(path):
+  local_resource("jest", "", serve_cmd="cd %s && yarn run jest --bail" % path)
+```
+
+Now Tilt is running your tests, how cool is that?
+
+## Conclusion
+Whether it's a small thing like how to get logs out of Kubernetes or a big thing like how to efficiently run tests, we think that sharing these patterns through Tilt Extensions will unlock step-level increases in engineering productivity. Just how NPM packages enabled people to do more things faster with JavaScript, we think Tilt Extensions will enable people to make their own developer experience more productive, faster.
```
landism: The future tense here gets tricky...it definitely goes back to making it feel like an announcement. Simply changing it to present tense might feel bad since extensions haven't actually delivered on promises yet? Though I think for these particular claims we can just drop the "will" since they're also both also qualified by "we think" and "enable", rather than something super concrete like "EXTENSIONS MAKE PEOPLE FASTER"

ellenkorbes:
I basically agree with all of Matt's outstanding comments.

docs/writing_tilt_extensions.md
```diff
@@ -0,0 +1,54 @@
+---
+title: Writing Tilt Extensions (Preview)
+layout: docs
+---
+Extensions are a simple way to bundle up Tiltfile functionality that should be shared with the world.
+
+Lots of Tilt users have implemented useful Tiltfile functionality that'd be great for others to be able to use as well. We've seen functions that force resources in to different namespaces, inject sidecars or even run tests. We created extensions to make it easy for Tiltfile authors to share those ideas, and for new Tiltfile authors to stand on the shoulders of giants by building on top of them.
```
ellenkorbes: s/in to/into?

PR #437, opened 2020-02-22
Link to Windows support issue on install page

victorwuky:


victorwuky:
@nicks : Does this work?

PR #444, opened 2020-02-27
Revert "tiltfile: all jekyll services should block on the 'make api' job (#393)"

maiamcc:
@nicks correct me if I'm wrong, but my understanding of #393 is that
`make-api` spits out changed files that will trigger other resources to
rebuild, so it's more efficient to block on `make-api`, right? (rather than
all resources need the results of make-api to run RIGHT.)

if that's the reasoning behind #393, then I believe it's rendered obsolete by
[tilt#2854](https://github.com/windmilleng/tilt/pull/2854) -- all resources automatically
block on `make-api` because it's a local resource, and there's no need to
specify it as a dep.

(The annoying thing about `make-api` being a resource dep for everything is
that `tilt up blog-site` pulls in `make-api` as a dep, and afaict it's
unnecessary and slow--that's why I want to revert.)

Does that all sound right?


nicks:
no, that doesn't sound right to me. https://github.com/windmilleng/tilt/pull/2854 ensures that local_resource doesn't run in parallel with other commands, but doesn't say anything about oder

maiamcc:
Local resources are already [guaranteed to come first](https://github.com/windmilleng/tilt/blob/7dc69dbde8204c70dc227818116d344121befcff/internal/engine/buildcontrol/build_control.go#L113) unless they depend on some other non-local resource

nicks:
is that something we guarantee in public-facing documentation? I thought it was an internal implementation detail, and not something we want people relying on

nicks:
(to be clear, I'm fine with this if we decide that we do want to make this guarantee, i just want to make sure it's documented somewhere)

maiamcc:
Fair--how's this as a solution (so that running `tilt up blog-site` doesn't pull in `make-api` and make you wait)?

PR #446, opened 2020-02-28
docs: add an example/shell guide

nicks:
Hello @ellenkorbes, @maiamcc,

Please review the following commits I made in branch nicks/ch4920/shell:

4e6f0fccbaed7b68287442a1d692926e2490ab41 (2020-02-28 14:31:03 -0500)
docs: add an example/shell guide


nicks:
As much as possible, I'd like to try to get agreement on common structure and language across the different guides so we can just copy + paste + tweak things.

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+In this example, we're going to take you through a very simple shell script that
+serves static HTML.
 
-Oneup is a simple app that uses busybox to serve HTML files.
+We'll use Tilt to:
 
-First, check out the Tilt repo.
+- Run the server on Kubernetes
+- Measure how long it takes to deploy
+- Optimize the updates for fast feedback.
```
maiamcc: "the updates" is vague. Maybe like, "optimize the propagation of local changes up to the running server" or something?

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+```shell
+echo "Serving files on port 8000"
+busybox httpd -f -p 8000
 ```
-git clone https://github.com/windmilleng/tilt
-cd tilt/integration/oneup
+
+
+1) A [Dockerfile](https://github.com/windmilleng/tilt-example-html/blob/master/0-base/Dockerfile) that builds the image.
```
maiamcc: nit: be consistent with punctuation at the end of these bullets

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
 ```
 
-We'll be looking at the [oneup](https://github.com/windmilleng/tilt/tree/master/integration/oneup) project
-in `integration/oneup`.
+The first line tells Tilt to build an image with the name `example-html-image`
 
-In the oneup directory, run
+The second line tells Tilt to load the Kubernetes deployment yaml. The image name in the `docker_build`
+call must match the image name in the `example-html` Deployment.
```
maiamcc: if you wanna be even more specific, "must match the `spec.container.image` value"

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+
+The last line configures port-forwarding so that your server is
+reachable at http://localhost:8000/. The resource name in the `k8s_resource` call
+must match the deployment name in the Kubernetes yaml.
```
maiamcc: again, if you want to be more specific, "must match the Deployment's `metadata.name`"

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
 ```
+git clone https://github.com/windmilleng/tilt-example-html
+cd tilt-example-html/0-base
 tilt up
 ```
 
-Your terminal will turn into a status box that lets you watch your server come up. When it's ready,
-you will see the status icon turn green. The logs in the botton pane will display
-"Serving oneup on container port 8000."
+Your terminal will turn into a status box that lets you watch your server come
```
maiamcc: Do we want to encourage web UI here instead?

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+botton pane will display "Serving files on port 8000." The snapshot below
+lets you see what it should look like.
+
+[Snapshot](https://cloud.tilt.dev/snapshot/AejkyuULr2AjWu50Eck=)
+
+## Step 1: Let's Add Benchmark Trickery
+
+Before we try to make this faster, let's measure it.
+
+Tilt has two major escape hatches for running commands locally:
```
maiamcc: imo this doesn't differentiate super well btwn local and local_resource, i might just not even mention `local` cuz it doesn't seem hugely relevant

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+
+- `local(command)` which runs a local command for its output
+
+- `local_resource(name, command)` which lets you run local jobs to trigger new updates, or run local servers
+
+For more detail, see the [local resource guide](local_resource.html)
+
+We add a `local_resource` to our
+[Tiltfile](https://github.com/windmilleng/tilt-example-html/blob/master/1-measured/Tiltfile)
+that records when an update starts.
```
maiamcc:
```suggestion
that records when an update starts. We've also modified our server itself to read that start time and print the time elapsed.
```
Or something--otherwise it seems like magic/unclear how the local resource helps you measure deploy time

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+# When the contents of this directory change, example-html will automatically re-deploy.
+local_resource(
+  'deploy',
+  'date +%s > start-time.txt',
+  trigger_mode=TRIGGER_MODE_MANUAL)
+```
+
+The `local_resource()` call creates a local resource named `deploy`. The second
+argument is the script that it runs. The third argument adds a button to the
+Tilt UI, to re-start the deploy.
```
maiamcc: why is `TRIGGER_MODE_MANUAL` necessary here? it has no deps, it would never auto-run

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+  run('./report-deployment-time.sh'),
+])
 ```
-tilt down
+
+We've added a new parameter to `docker_build()` with two `live_update` steps.
+
+The first step syncs the code from the current directory (`.`) into the container at directory `/app`.
+
+The second step runs our script to report the deployment time. This is just for measuring!
```
maiamcc:
```suggestion
The second step runs our script to report the deployment time. (This is just for measuring!)
```

maiamcc:
(But also, I feel like we should print `done in XXX` in the logs for the WHOLE build pipeline like we do for the deployment step--if that's easy to implement, I'd say we should go for it because it'll make benchmarking services feel much nicer!)

ellenkorbes:
I didn't run this locally, so I might be wrong, but I feel this story could be a bit more visual, with more screenshots.

Also, I'd favor screenshots as opposed to snapshots (maybe we can keep the snapshot links in parenthesis or something), since snapshots are quite busy and might be confusing to beginners.

I'm open to being wrong :-)

nicks:
Let me just address the high-level comments up-front:

re: durations - I don't think Sidebar / Log build times are what we want here. Those include the time to build the image + apply the yaml. They do NOT including pod scheduling/init time. This will be much more obvious with other languages. It's trivial for busybox.

I think there's a longer conversation we need to have about what different durations in the UI mean and how to communicate that effectively. I expect this to evolve over time. It's not the kind of complexity I think we want to have in the

re: benchmarking - I think it's still valuable to demonstrate how to build your own benchmarking, just for the educational value. But maybe that's a whole separate discussion. Are there other local_resource things you'd rather do?

re: screenshots - totally agree. I'm going to replace them with clickable screenshots that lead to snapshots. This makes me wonder if we should make a simple version of the snapshot embeddable - like a gist - but obviously don't want to open that box now

maiamcc:
> re: durations - I don't think Sidebar / Log build times are what we want here. Those include the time to build the image + apply the yaml. They do NOT including pod scheduling/init time. This will be much more obvious with other languages. It's trivial for busybox.

Yep you're right, I wasn't thinking. Fridayyyy. 

> re: benchmarking - I think it's still valuable to demonstrate how to build your own benchmarking, just for the educational value. But maybe that's a whole separate discussion. Are there other local_resource things you'd rather do?

Nah, this seems like a perfectly fine way to benchmark, if we're gonna benchmark (and you make a persuasive point, so i'm happy to go with it). I wonder if it's worth calling this out, e.g. "the tilt logs show you how long it takes to build+deploy your app, but don't take into account pod startup time" or something. I don't feel strongly tho.

docs/example_static_html.md
maiamcc: channeling DB here but: I don't think "deploy" is the word we want. "Measure how long it take for local edits to propagate to the server"?

...I'm being verbose cuz I want to be precise, and like I mention below, "updates" is a liiiiittle vague/overloaded... but. hmm. "Measure how long it takes to update our server with new code"? "Measure how long it takes to see new code running in the cluster"?

docs/example_static_html.md
```diff
@@ -3,50 +3,190 @@ title: "Example: Plain Old Static HTML"
...
+[Tiltfile](https://github.com/windmilleng/tilt-example-html/blob/master/1-measured/Tiltfile)
+that records when an update starts.
+
+```python
+k8s_resource('example-html', port_forwards=8000, resource_deps=['deploy'])
+
+# Records the start time.
+# When the contents of this directory change, example-html will automatically re-deploy.
+local_resource(
+  'deploy',
```
maiamcc: again, idk if "deploy" is right here (cuz later this resource will trigger a live_update)--though also I don't think "update" is quite right either. "re-up app"? bluh.

ellenkorbes:
Just as a suggestion, I like to phrase it as "time from code change to new
process."

docs/example_static_html.md
nicks: ya, I'm not sure. I think "deploy a code change" is probably ok even when using live update? let's revisit this if people are confused about it.

docs/example_static_html.md
nicks: good point!

docs/example_static_html.md
nicks: i just removed this

docs/example_static_html.md
maiamcc: mmmm okay yeah fair.

docs/example_static_html.md
maiamcc: Consider reversing these to encourage use of the web UI as primary:

> Tilt will open a browser showing the web UI, a unified view that shows you app status and logs. Your terminal will also turn into a status box if you'd like to watch your server come up there.

(or something)

docs/example_static_html.md
```diff
@@ -3,50 +3,176 @@ title: "Example: Plain Old Static HTML"
...
-Your terminal will turn into a status box that lets you watch your server come up. When it's ready,
-you will see the status icon turn green. The logs in the botton pane will display
-"Serving oneup on container port 8000."
+Your terminal will turn into a status box that lets you watch your server come
+up. Tilt will open a browser view of the same status.
+
+When it's ready, you will see the status icon turn green. The logs in the
+botton pane will display "Serving files on port 8000."
+
+[![Step 0](assets/docimg/example-static-html-image-1.png)](https://cloud.tilt.dev/snapshot/AejkyuULr2AjWu50Eck=){:.is-image}
```
maiamcc: consider adding a caption (with or without descriptive information like "the server is up!") that includes "click the image to see an interactive snapshot"--cool change to show off snapshot functionality, and users might not think to click otherwise.

docs/example_static_html.md
```diff
@@ -3,50 +3,176 @@ title: "Example: Plain Old Static HTML"
...
 ```
 
-Your terminal will turn into a status box that lets you watch your server come up. When it's ready,
-you will see the status icon turn green. The logs in the botton pane will display
-"Serving oneup on container port 8000."
+Your terminal will turn into a status box that lets you watch your server come
+up. Tilt will open a browser view of the same status.
+
+When it's ready, you will see the status icon turn green. The logs in the
+botton pane will display "Serving files on port 8000."
```
maiamcc: also, are you gonna tell them to navigate to `localhost:8000`? (you can mention it's clickable from the UI or not, whatever)

docs/example_static_html.md
```diff
@@ -3,50 +3,176 @@ title: "Example: Plain Old Static HTML"
...
+# Records the time from a code change to a new process.
+# Normally, you would let Tilt do deploys automatically, but this
+# shows you how to set up a custom workflow that measures it.
+local_resource(
+  'deploy',
+  'date +%s > start-time.txt')
 ```
-tilt down
+
+The `local_resource()` call creates a local resource named `deploy`. The second
```
maiamcc: could cut out this line and notate this in code, if you want--seems a bit more concise to me:
```
local_resource(
  'deploy',  # name o the local resource
  'date %s...'  # command to run
)

File: docs/example_static_html.md [hunk 14]
```diff
@@ -3,50 +3,176 @@ title: "Example: Plain Old Static HTML"
...
+argument is the script that it runs.
+
+Let's click the deploy button and see what happens!
```
maiamcc: is the button subtle enough that you should elaborate "the button next to the 'deploy' resource" or show a screenshot?

docs/example_static_html.md
```diff
@@ -3,50 +3,176 @@ title: "Example: Plain Old Static HTML"
...
+
+[Recommended Structure](https://github.com/windmilleng/tilt-example-html/blob/master/2-recommended)
+
+Obviously, our busybox example is very silly. We just wanted to show you how
+Tilt can work with any language, even a silly one.
+
+Other examples:
+
+<ul>
+  {% for page in site.data.examples %}
```
maiamcc: nice!

docs/example_static_html.md
```diff
@@ -3,50 +3,176 @@ title: "Example: Plain Old Static HTML"
...
-</div>
+## Step 1: Let's Add Benchmark Trickery
 
-Type `b` to open `oneup` in a browser window.
-Your browser will open `http://localhost:8100`.
-You should see the text `🍄 One-Up! 🍄`.
+Before we try to make this faster, let's measure it.
 
-Congratulations! You've run your first server with `tilt`.
+Tilt has escape hatches for running commands locally, so that you can integrate your existing scripts.
```
maiamcc: Even this feels like a red herring, because we're not using it to integrate with an existing script--I might just take this out.

docs/example_static_html.md
nicks: done

docs/example_static_html.md
nicks: done

docs/example_static_html.md
nicks: eh, seems unnecessary, we can revisit if any users are confused about what "port 8000" means

docs/example_static_html.md
nicks: done

docs/example_static_html.md
nicks: The problem with inline comments is that they end up wrapping weirdly on mobile

docs/example_static_html.md
nicks: done

PR #451, opened 2020-03-02
Fix force-update-button.png in blog post

victorwuky:
Right now, the image is broken, because the location of it is incorrect. This PR adds the image to the right directory and updates the markdown file for the image location, to be consistent with convention / formatting of newer blog posts.

See blog post at https://blog.tilt.dev/2019/11/15/local-resource.html

| Before | After |
| --- | --- |
| Screen Shot 2020-03-02 at 10 52 13 AM | Screen Shot 2020-03-02 at 12 28 18 PM |


victorwuky:
Thanks for reviewing and approving @nicks

PR #452, opened 2020-03-02
docs: add docs for os.getcwd()

nicks:
Hello @landism, @maiamcc,

Please review the following commits I made in branch nicks/osgetcwd:

6af87c4251be456b237a95d254ae526361fa73ed (2020-03-02 13:19:47 -0500)
docs: add docs for os.getcwd()

### DIFF
--- a/api/modules/os.py
+++ b/api/modules/os.py
@@ -15,3 +15,14 @@ as Python dictionaries, including:
 
 See the `Starlark spec <https://github.com/bazelbuild/starlark/blob/master/spec.md#built-in-methods>`_ for more.
 """
+
+def getcwd() -> str:
+  """Returns a string representation of the current working directory.
+
+  The current working directory is the directory containing the currently executing Tiltfile.
+  If your Tiltfile runs any commands, they run from this directory.
+
+  While calling :meth:load or :meth:include to execute another Tiltfile,
+  returns the directory of the loaded/included Tiltfile.
+  """
+  pass
--- a/src/_data/api/functions.yaml
+++ b/src/_data/api/functions.yaml
@@ -43,6 +43,7 @@ functions:
 - version_settings
 - watch_file
 - workload_to_resource_function
+- os.getcwd
 - config.define_string_list
 - config.parse
 - config.set_enabled_resources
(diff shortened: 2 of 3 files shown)


api/modules/os.py
```diff
@@ -15,3 +15,13 @@
 
 See the `Starlark spec <https://github.com/bazelbuild/starlark/blob/master/spec.md#built-in-methods>`_ for more.
 """
+
+def getcwd() -> str:
+  """Returns a string representation of the current working directory.
+
+  If your Tiltfile runs any commands, they will inherit this working directory.
+
+  When your Tiltfile loads() another Tiltfile, the working directory will
```
landism: I see a couple potential ambiguities here:
1. This sounds kind of like a state change, like it might set the expectation that `load('foo/bar'); print(os.getcwd())` would return 'foo/bar' rather than '.' (or whatever).
2. (I'm assuming) this returns the Tiltfile's directory, not the CWD of the calling process.

Maybe something like this?

> The working directory for the currently executing Tiltfile. If your Tiltfile runs any commands, they run from this directory.
> If this is called while `:meth:load`ing or `:meth:include`ing a Tiltfile, it will return the directory of the loaded/included Tiltfile.

or maybe #1 above is mostly addressed by changing "When" to "While".

nicks: done! tweaked your language a bit to match the verb tenses in the rest of the doc

api/modules/os.py
```diff
@@ -15,3 +15,14 @@
...
 See the `Starlark spec <https://github.com/bazelbuild/starlark/blob/master/spec.md#built-in-methods>`_ for more.
 """
+
+def getcwd() -> str:
+  """Returns a string representation of the current working directory.
+
+  The current working directory is the directory containing the currently executing Tiltfile.
+  If your Tiltfile runs any commands, they run from this directory.
+
+  While calling :meth:load or :meth:include to execute another Tiltfile,
```
maiamcc: Hmm, grammar seems awks? Maybe:

> If called in Tiltfile B, which is pulled into Tiltfile A via :meth:load or :meth:include, returns the directory of Tiltfile B.

maiamcc:
Lol whoops too late 😅 my comment was small, this is fine as-is!

PR #458, opened 2020-03-05
guides: python + flask server [ch4825]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/python-guide:

ff0638434cc239d3008d2dd02f6129ba07dc0014 (2020-03-05 18:29:18 -0500)
guides: python + flask server [ch4825]


docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+We'll use Tilt to:
 
-[Servantes](https://github.com/windmilleng/servantes) is a personalized homepage
-app implemented with many microservices in multiple languages. You can run it
-with the commands
+- Run the server on Kubernetes
+- Measure the time from a code change to a new process
+- Optimize that time for fast feedback
 
+Obviously, this is a silly example. But it can be a useful example to confirm that Tilt is working
```
nicks: Let's axe this line, it was specific to busybox. Flask isn't silly at all!

docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+Can we do better?
+
+## Step 2: Let's Optimize It
+
+When we make a change to a file, we currently have to build an image, deploy new Kubernetes configs,
+and wait for Kubernetes to schedule the pod.
+
+With Tilt, we can skip all of these steps, and instead [live-update](https://docs.tilt.dev/live_update_tutorial.html) the pod in place.
+
+[Our new Tiltfile](https://github.com/windmilleng/tilt-example-html/blob/master/2-recommended/Tiltfile) contains the following new code:
```
nicks: this is pointing to the HTML example?

docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+To start this server on Kubernetes, we need 3 configs:
+
+1. A [Dockerfile](https://github.com/windmilleng/tilt-example-python/blob/master/0-base/Dockerfile) that builds the image
+2. A [Kubernetes deployment](https://github.com/windmilleng/tilt-example-python/blob/master/0-base/kubernetes.yaml) that runs the image
+3. And finally, a [Tiltfile](https://github.com/windmilleng/tilt-example-python/blob/master/0-base/Tiltfile) that ties them together:
+
+```python
+docker_build('example-python-image', '.')
+k8s_yaml('kubernetes.yaml')
+k8s_resource('example-python', port_forwards=5000)
```
nicks: can we use the same application port for all our examples?

We usually recommend people use port 5000 for the registry, see:
https://docs.docker.com/registry/deploying/
https://github.com/windmilleng/kind-local/blob/master/kind-with-registry.sh

and this example will not work for them as-is

docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+```
+
+Tilt will open a browser showing the web UI, a unified view that shows you app
+status and logs. Your terminal will also turn into a status box if you'd like to
+watch your server come up there.
+
+When the server is ready, you will see the status icon turn green. The log pane will display some output from Flask, starting with:
+> Serving Flask app "app"
+
+[![The server is up! (Click the screenshot to see an interactive snapshot.)](assets/docimg/example-python-1-serverup.png)](https://cloud.tilt.dev/snapshot/AZSXjOYLDi6w9IoCnCI=){:.is-image}
```
nicks: can you change this to use figure/figcaption like in the other example?

docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+[![Result of clicking the button  on the “deploy” resource](assets/docimg/example-python-3-naive-deploy.png)](https://cloud.tilt.dev/snapshot/AcyUjOYLaeKVxQ1XZjk=){:.is-image}
+*Clicking the button triggers the "deploy" local_resource, which in turn kicks off an update to the server. (Click the screenshot to see an interactive view.)*
+
+| Approach | Deploy Time (after initial)
+|---|---|
+| Naive | 10-11s |
+
+Can we do better?
+
+## Step 2: Let's Optimize It
```
nicks: the outline here:
https://github.com/windmilleng/tilt.specs/blob/master/outboarding_guide_outline.md

discussed having a separate step for optimizing the dockerfile? I think you still need that here. When I tried running this example, the first 'tilt up' was not great because `pip install` runs every time something changes

docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+    # if all that changed was start-time.txt, make sure the server
+    # reloads so that it will reflect the new startup time
+    run('touch /app/app.py', trigger='start-time.txt'),
+
+    # add a congrats message!
+    run('sed -i "s/Hello cats!/{}/g" /app/templates/index.html'.format(congrats)),
+])
+```
+
+We've added a new parameter to `docker_build()` with four `live_update` steps that are, in order:
```
nicks: I also saw your example code add debug=True to the Flask app. is that worth calling out here, to make it clear that Flask is doing the live-reloading?

Though TBH, my understanding of Flask is that you're supposed to use `flask run` when you're in dev mode. I think we would recommend that people use entrypoint='flask run' in their Tiltfile, rather than hard-coding debug mode in the app, right?

docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+### Final Score
+
+| Approach | Deploy Time (after initial)
+|---|---|
+| Naive | 10-11s |
+| With live_update | 1-2s |
+
+You can try the server here:
+
+[Recommended Structure](https://github.com/windmilleng/tilt-example-python/blob/master/2-live_update)
```
nicks: I think it would be good to keep a convention of 'N-recommended' being the recommended workflow

docs/example_python.md
maiamcc: I mean, it's a silly hello world server -- but, fair

docs/example_python.md
```diff
@@ -1,30 +1,194 @@
...
+status and logs. Your terminal will also turn into a status box if you'd like to
+watch your server come up there.
+
+When the server is ready, you will see the status icon turn green. The log pane will display some output from Flask, starting with:
+> Serving Flask app "app"
+
+[![The server is up! (Click the screenshot to see an interactive snapshot.)](assets/docimg/example-python-1-serverup.png)](https://cloud.tilt.dev/snapshot/AZSXjOYLDi6w9IoCnCI=){:.is-image}
+*The server is up! (Click the screenshot to see an interactive view.)*
+
+## Step 1: Let's Add Benchmark Trickery
```
ellenkorbes: This section left me a bit lost because it described the resource that created the `start-time.txt` file, but it didn't tell me what was it that then read it to calculate the time elapsed. IMO it would be helpful to have a sentence or two to mention that `get_update_time_secs()` was added.

docs/example_python.md
ellenkorbes: Can confirm, just ran into this.

docs/example_python.md
maiamcc: whoops ty

docs/example_python.md
maiamcc: done

docs/example_python.md
maiamcc: done

docs/example_python.md
maiamcc: changed -- better?

docs/example_python.md
maiamcc: ooh good call, let me look into that.

docs/example_python.md
maiamcc: yep yep you're totally right

maiamcc:
Really good feedback, thanks all. I propose merging this with the current structure (tho feel free to make any more low-level comments you have), and in a subsequent PR addressing:
- best practice for flask apps in re: how to invoke them in prod vs. dev (i.e. entrypoint `flask run` instead of hardcoding debug mode)
- another step to optimize the dockerfile

docs/example_python.md
```diff
@@ -1,30 +1,225 @@
...
+Obviously, this particular server doesn't do much, but it's still a useful example to confirm that Tilt is working
+as expected in your environment.
+
+All the code is in this repo:
+
+[tilt-example-python](https://github.com/windmilleng/tilt-example-python)
+
+To skip straight to the fully optimized setup, go to this subdirectory:
+
+[Recommended Tiltfile](https://github.com/windmilleng/tilt-example-python/tree/master/2-live_update)
```
nicks: i think this is at 2-recommended now?

docs/example_python.md
ellenkorbes: Perfect, thanks!

docs/example_python.md
maiamcc: whoooops yes.

PR #462, opened 2020-03-10
Improve using Tilt extensions docs

victorwuky:
Goals of this PR:

- Copyedits and general clarifications.
- Best practice of not changing extensions locally.
- Inform that versioning is not supported.

### DIFF
--- a/docs/using_tilt_extensions.md
+++ b/docs/using_tilt_extensions.md
@@ -23,7 +23,7 @@ The `tilt_modules` directory should be committed to your repo (**not** `gitignor
 
 ## Changing an Extension
 
-We don't recommend changing extensions directly in `tilt_modules`. If you want to customize an extension you are already using, consider [contributing a change or a brand new extension](writing_tilt_extensions.html). Once the change is merged into [windmilleng/tilt-extensions](https://github.com/windmilleng/tilt-extensions/), get the updated or new extension (first deleting your local version in `tilt_modules` as necessary). Commit your changes to source control so that they will propagate to your team.
+We don't recommend changing extensions directly in `tilt_modules`. If you want to customize an extension you are already using, consider [contributing a new extension](writing_tilt_extensions.html).
 
 Tilt doesn't support versioning of extensions currently. If you are interested in versioning, [let us know](https://tilt.dev/contact).
 
--- a/docs/using_tilt_extensions.md
+++ b/docs/using_tilt_extensions.md
@@ -28,4 +28,4 @@ We don't recommend changing extensions directly in `tilt_modules`. If you want t
 Tilt doesn't support versioning of extensions currently. If you are interested in versioning, [let us know](https://tilt.dev/contact).
 
 ## Next Steps
-If you have an idea for a new extension, [give us a shout](https://tilt.dev/contact). If you'd like to create a new extension, visit [Writing Tilt Extensions](writing_tilt_extensions.html).
+If you have an extension request, [give us a shout](https://tilt.dev/contact). If you'd like to contribute a new extension, visit [Writing Tilt Extensions](writing_tilt_extensions.html).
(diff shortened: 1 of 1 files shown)


victorwuky:
@landism : Can you please review? Thanks!

docs/using_tilt_extensions.md
```diff
@@ -15,9 +15,17 @@ Once you've found an extension, using it couldn't be easier! Say we wanted to us
...
+
+`hi` is the name of the function defined in [hello_world/Tiltfile](https://github.com/windmilleng/tilt-extensions/blob/master/hello_world/Tiltfile) and now made available to your local project Tiltfile. Simply call `hi()` anywhere to execute it, printing "Hello world!".
 
 The `tilt_modules` directory should be committed to your repo (**not** `gitignore`'d) to ensure that your teammates don't have to download it in the future, and to ensure that everyone on your team is using the same version of the extension.
 
+## Changing an Extension
+
+We don't recommend changing extensions directly in `tilt_modules`. If you want to customize an extension you are already using, consider [contributing a new extension](writing_tilt_extensions.html).
+
+Tilt doesn't support versioning of extensions currently. If you are interested in versioning, [let us know](https://tilt.dev/contact).
```
landism: This doesn't seem like it fits under "Changing an Extension". How about at the end of "Using an Extension"?

docs/using_tilt_extensions.md
```diff
@@ -15,9 +15,17 @@ Once you've found an extension, using it couldn't be easier! Say we wanted to us
...
+`ext://hello_world` is resolved to https://github.com/windmilleng/tilt-extensions/blob/master/hello_world/Tiltfile by Tilt. When the extension is loaded into your project for the first time, Tilt copies the remote Tiltfile into the `tilt_modules` directory of local your project. So in this case Tilt would write the contents of `ext://hello_world` to `tilt_modules/hello_world`.
+
+`hi` is the name of the function defined in [hello_world/Tiltfile](https://github.com/windmilleng/tilt-extensions/blob/master/hello_world/Tiltfile) and now made available to your local project Tiltfile. Simply call `hi()` anywhere to execute it, printing "Hello world!".
 
 The `tilt_modules` directory should be committed to your repo (**not** `gitignore`'d) to ensure that your teammates don't have to download it in the future, and to ensure that everyone on your team is using the same version of the extension.
```
landism: I think something along these lines would be good: "Once Tilt has downloaded an extension, it will not update it, so you won't get surprise breakages. We will soon be implementing support for intentionally updating its local copies of extensions."

(or adjust the "soon" depending on where it might be triaged, but it should be on the list, even if just to define a manual process)

landism: or maybe that's out of scope of this PR? In which case, FFTI and I can post a separate PR.  Mostly bringing it up because it was the main thing I was hoping for out of the "we should document lack of versioning" conversation yesterday.

docs/using_tilt_extensions.md
victorwuky: Makes sense. Moved to end of "Using an Extension".

docs/using_tilt_extensions.md
victorwuky: What is "FFTI"?

landism: Feel Free To Ignore

victorwuky: I added your first sentence, but not the second sentence. Let's add the second sentence, but not use "soon", and instead link to a GitHub issue where we can use that as the source of truth of expressing "soon-ness". 

I figure we can add the second sentence once we actually get the feature out the door and it is more public. So maybe it would be a bit weird to have an open GitHub issue that references a non-advertised feature right now? (I don't think it _really_ matters, since this very discussion is in a public repo. But just to keep the https://github.com/windmilleng/tilt issue tracker somewhat clean I think would make sense.)

PR #464, opened 2020-03-11
docs: example go project

nicks:
Hello @ellenkorbes, @maiamcc,

Please review the following commits I made in branch nicks/ch4750/go:

47365761a2fb7d3e876772f40073bb4954dbd670 (2020-03-11 14:23:34 -0400)
docs: example go project


docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
-We here you! We're working on a better guide for this.
+Kubernetes is a huge wrench in the works.
 
-For now, check out:
+Let's fix this.
 
-- [abc123](https://github.com/windmilleng/abc123) a mini microservice app with a Go server called `fe`.
-- The [Live Update Tutorial](live_update_tutorial.html), which optimizes a Go server.
-- Servantes (below), our microservice app in multiple languages.
+In this example, we're going to take you through a very simple shell script that
```
ellenkorbes: Not a shell script.

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+serves static HTML.
+
+We'll use Tilt to:
+
+- Run the server on Kubernetes
+- Measure the time from a code change to a new process
+- Optimize that time for fast feedback
+
+Obviously, this is a silly example. But it can be a useful example to confirm that Tilt is working
```
ellenkorbes: https://github.com/windmilleng/tilt.build/pull/458#discussion_r389059035

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+[tilt-example-go](https://github.com/windmilleng/tilt-example-go)
+
+To skip straight to the fully optimized setup, go to this subdirectory:
+
+[Recommended Tiltfile](https://github.com/windmilleng/tilt-example-go/blob/master/3-recommended/Tiltfile)
+
+## Step 0: The Simplest Deployment
 
-## Servantes
+Our server is a simple HTML server that uses Gorilla Mux for routing requests:
```
ellenkorbes: The package name is usually written as `gorilla/mux`; I couldn't find any instance of "Gorilla Mux." (https://github.com/gorilla/mux)

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+
+```python
+docker_build('example-go-image', '.', 
+    dockerfile='deployments/Dockerfile')
+k8s_yaml('deployments/kubernetes.yaml')
+k8s_resource('example-go', port_forwards=8000)
 ```
 
-Each widget is implemented by a different microservice backend.
+The first line tells Tilt to build an image with the name `example-html-image`
```
ellenkorbes: s/`example-html-image`/`example-go-image`

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+
+We've added a `live_update` parameter to `docker_build()` with two `sync` steps.
+They copy the code from the `./build` and `./web` directories into the container.
+
+We've also added a new parameter: `entrypoint="find . | entr -r ./build/tilt-example-go"`.
+
+`entr` is a tool that automatically restarts a shell command whenever the watched
+file changes. This command restarts our server every time the pod is updated.
+
+Congratulations on finishing this guide!
```
ellenkorbes: This makes me wanna close the tab. Let's hold it until after the final score.

ellenkorbes:
A few possibly-too-opinionated observations:

1) We're serving two files; there's no reason to use `gorilla/mux` here. Especially given how the Go community is averse to importing random things, and has a preference for writing it by hand unless there's very good reason not to... `gorilla/mux` just feels out of place. 

We can achieve all we need with a handful of lines of standard library instead.

2) The way we're doing record-start-time.sh adds a layer of indirection that takes too much mental effort to unravel. It took me a whole 3 minutes and I knew what I was looking for—IMO that's too much complexity for an on-boarding guide. 

Ideally users should be able to understand every file at a glimpse so that they can focus on learning how Tilt works, not trying to understand what the code is doing. When we have a shell script that generates a file that's part of a package that's then imported by main app... that's too much.

3) IMO the issue raised in yesterday's meeting was valid and we should address why the benchmark reports one number and Tilt reports another one. A single sentence explaining that they're measuring different things would do it.

(I haven't slept in many years so if I'm tripping feel free to disregard everything.)

nicks:
Let me address the high-level questions first:

1) I don't have any strong attachment to gorilla/mux. I do think it's important that any guide we have show how third-party dependencies work in Tilt. "Just use the standard library" would be fine advice in a Go tutorial, I think it's unhelpful in a tutorial on how to setup Go projects with Tilt.

Is there a package you'd prefer we show instead? [windmilleng/fast](https://github.com/windmilleng/fast/blob/master/1-vendor/main.go) uses Cobra for example -- I'd be fine with standardizing on "all examples should use a third-party flag parsing library

2) I heard a few different objections in your comment here, but I'm not sure what the relative priority is, or what would resolve them. Is it just that you want the generated start.go to live in the same package as main.go?

3) Sure, I'm fine with adding some text to explain this, I opened a ticket to agree on some language, and update the guides: https://app.clubhouse.io/windmill/story/5311/explain-why-onboarding-guide-metrics-differ-from-tilt-metrics

nicks:
sent as https://github.com/windmilleng/tilt.specs/pull/27, lemme know how that looks?

ellenkorbes:
1. 

> show how third-party dependencies work in Tilt

But we're not doing that. We're not showing anything related to third party dependencies. If we remove `gorilla/mux` from the code, nothing of what we're explaining will change.

It's like the Chekov's gun principle in theater: "If in the first act you have hung a pistol on the wall, then in the following one it should be fired. Otherwise don't put it there." 

If we'd like to talk about third party dependencies and bring it to fore, fine, but right now I feel it's just taking attention away from what is actually the main point—how to use a Tiltfile to tie together a bunch of different components.

2. That would help a lot, yes.

3. YES! Perfect! 👏

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+}
 ```
-git clone https://github.com/windmilleng/servantes
-cd servantes; tilt up
+
+
+1) A [Dockerfile](https://github.com/windmilleng/tilt-example-go/blob/master/0-base/deployments/Dockerfile) that builds the image
+
+2) A [Kubernetes deployment](https://github.com/windmilleng/tilt-example-html/blob/master/0-base/deployments/kubernetes.yaml) that runs the image
```
maiamcc: wrong repo

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+
+3) And finally, a Tiltfile that ties them together
```
maiamcc: `Tiltfile` should have a link?

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+    <img src="assets/docimg/example-go-image-1.png">
+  </a>
+  <figcaption>The server is up! Click the screenshot to see an interactive snapshot.</figcaption>
+</figure>
+
+## Step 1: Let's Add Benchmark Trickery
+
+Before we try to make this faster, let's measure it.
+
+Tilt can run commands locally, so that you can integrate your existing
```
maiamcc: replace with [new copy from the template](https://github.com/windmilleng/tilt.specs/pull/27#pullrequestreview-373643423)

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+local_resource(
+    'deploy',
+    './record-start-time.sh',
+)
+```
+
+The `local_resource()` call creates a local resource named `deploy`. The second
+argument is the script that it runs.
+
+Let's click the button on the `deploy` resource and see what happens!
```
maiamcc: not a hard requirement, but in my guide I put a screenshot of the button, cuz I think it's not immediately obvious?

example-python-2-deploy-button

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+
+<figure>
+  <a class="is-image" href="https://cloud.tilt.dev/snapshot/AbakyOYLWWvW6J2BJFo=">
+    <img src="assets/docimg/example-static-html-image-2.png">
+  </a>
+  <figcaption>Step 1 complete. Click the screenshot to see an interactive snapshot.</figcaption>
+</figure>
+
+| Approach | Deploy Time |
```
maiamcc: as per #465, this should be explicit html so we can style it (unless you reject the premise of #465 in which case 🤷‍♀ )

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+|---|---|
+| Naive | 4.2s |
+
+Can we do better?
+
+## Step 2: Let's Optimize for the Go Compiler
+
+When we make a change to a file, we currently copy all the source files to a
```
maiamcc:
```suggestion
Currently, when we make a change to a file, we copy all the source files to a
```

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+
+
+## Step 3: Let's Optimize It
+
+When we make a change to a file, we currently have to build an image, deploy new Kubernetes configs,
+and wait for Kubernetes to schedule the pod.
+
+With Tilt, we can skip all of these steps, live-updating the pod in place.
+
+Here's our [new Tiltfile](https://github.com/windmilleng/tilt-example-html/blob/master/3-recommended/Tiltfile)
```
maiamcc: wrong link

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+fast. How can we better use the Go tools how they're meant to be used?
+
+With `local_resource`, we can compile the Go binary locally, and copy the binary to
+the container.
+
+Here's our [new Tiltfile](https://github.com/windmilleng/tilt-example-go/blob/master/2-optimized/Tiltfile) 
+with the following new code:
+
+```python
+local_resource(
```
maiamcc: shouldn't this have a resource_dep on `deploy` (and the `docker_build` have a resource_dep on this local)?

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+  only=[
+    './build',
+    './web',
+  ])
+```
+
+We've added a `local_resource()` that compiles the Go binary locally for our
+Linux container.
+
+We've added a new `only` parameter to `docker_build()`. This makes sure that the
```
maiamcc:
```suggestion
We've also added a new `only` parameter to `docker_build()`. This makes sure that the
```

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+only directory in the Docker build context are the `./build` directory (for the
+compiled binary) and the `./web` directory (for the template and image files).
```
maiamcc: maybe worth explaining WHY we're doing this: to avoid stepping on `example-go-compile`'s toes. (if you're feeling really hardcore, i bet a diagram would be illuminating to many readers)

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+  ],
+  entrypoint = 'find . | entr -r ./build/tilt-example-go')
+```
+
+We've added a `live_update` parameter to `docker_build()` with two `sync` steps.
+They copy the code from the `./build` and `./web` directories into the container.
+
+We've also added a new parameter: `entrypoint="find . | entr -r ./build/tilt-example-go"`.
+
+`entr` is a tool that automatically restarts a shell command whenever the watched
```
maiamcc: oooh, slick workaround!

docs/example_go.md
```diff
@@ -3,29 +3,273 @@ title: "Example: Go"
...
+  <figcaption>Step 2 complete. Click the screenshot to see an interactive snapshot.</figcaption>
+</figure>
+
+| Approach | Deploy Time |
+|---|---|
+| Naive | 4.2s |
+| Local Compile | 3.5s |
+
+
+## Step 3: Let's Optimize It
```
maiamcc: i'm curious why you did local compile first, THEN live update, can you explain? (i don't think it's bad, it's just different than how we've organized things in the past)

nicks:
@ellenkorbes what if we explicitly called out the fact that Go automatically downloads dependencies in the container, and that's contributing to container build time. would that resolve the issue?

ellenkorbes:
It would justify mux's being there, yes.

docs/example_go.md
nicks: done

docs/example_go.md
nicks: added some text, tho i think we disagree on the reason

docs/example_go.md
nicks: good question! mainly following the outline. But I can think of a few reasons to do it this way:

- Adding `live_update` is a good capstone for a Tilt tutorial

- A lot of people in the Go community have started using this project for Go builds: 

https://github.com/thockin/go-build-template

which uses this Dockerfile:

https://github.com/thockin/go-build-template/blob/master/Dockerfile.in

which is consistent with the approach in this tutorial, so there's some prior art

docs/example_go.md
nicks: done

docs/example_go.md
nicks: I stole it from @ellenkorbes :robot:

docs/example_go.md
nicks: hahaha, done

docs/example_go.md
```diff
@@ -3,29 +3,303 @@ title: "Example: Go"
...
+k8s_yaml('deployments/kubernetes.yaml')
+k8s_resource('example-go', port_forwards=8000)
+```
+
+The first line tells Tilt to build an image with the name `example-go-image`
+in the directory `.` (the current directory).
+
+The second line tells Tilt to load the Kubernetes
+[Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment)
+yaml. The image name in the `docker_build` call must match the container `image`
```
ellenkorbes: Capitalize YAML?

docs/example_go.md
```diff
@@ -3,29 +3,303 @@ title: "Example: Go"
...
+cd tilt-example-go/0-base
+tilt up
+```
+
+Tilt will open a browser showing the web UI, a unified view that shows you app
+status and logs. Your terminal will also turn into a status box if you'd like to
+watch your server come up there.
+
+When it's ready, you will see the status icon turn green. The logs in the
+botton pane will display "Serving files on port 8000."
```
ellenkorbes: s/botton/bottom

docs/example_go.md
```diff
@@ -3,29 +3,303 @@ title: "Example: Go"
...
+| Naive | 4.2s |
+{:.benchmark-report}
 
+If you look closely, the elapsed time displayed in the Tilt sidebar is different
+than the benchmark our app logged. That's OK! In multi-service development,
+there are many benchmarks we care about -- the time to build the image, the time
+to schedule the process, and the time until the server is ready to serve
+traffic. 
+
+The Tilt sidebar gives you some default benchmarks, and the tools to capture
```
ellenkorbes: I'd italicize "and" here.

docs/example_go.md
maiamcc: that makes sense

docs/example_go.md
maiamcc: haha yep. so add `{:.benchmark-report}` after the markdown tables instead pls

docs/example_go.md
```diff
@@ -3,29 +3,303 @@ title: "Example: Go"
...
+  ])
 ```
-git clone https://github.com/windmilleng/servantes
-cd servantes; tilt up
+
+We've added a `local_resource()` that compiles the Go binary locally for our
+Linux container.
+
+We've also added a new `only` parameter to `docker_build()`.  We don't need the
+source files in the Docker build context. We only need the `./build` directory
```
maiamcc:
```suggestion
source files in the Docker build context; we only need the `./build` directory
```

maiamcc: I think you're right that we "disagree on the reason" that the `only` parameter needs explanation. Yes only pulling in the directories we need makes thing more efficient, but I'd rather talk explicitly about how this affects file-watching because that's the place I foresee people getting stuck.

Like, I might look at this and say "yeah it's a little more efficient to pull in only what we need, but doesn't seem like a huge deal, what happens if i leave this step out?" and then I see things double-building or building out of order; i'd be totally confused and there's nothing in the doc to help me understand what's happening.

PR #469, opened 2020-03-14
Clarify that yaml_filter uses regex-matching

phroggyy:
This updates the code comment for `yaml_filter` and adds a separate example of an exact name match.

Happy to make some modifications if someone thinks there's a clearer way to express this.


jazzdan:
Thanks @phroggyy!

PR #471, opened 2020-03-17
guides: add step to python guide for optimizing dockerfile

maiamcc:
if this looks right to you, i'll re-up the screenshots and snapshots
        to reflect the new output we expext to see


docs/example_python.md
```diff
@@ -171,21 +172,70 @@ your own benchmarks.
 
 Our benchmarks show this is slow. Can we do better?
 
-## Step 2: Let's Optimize It
+## Step 2: Why Is the Docker Build So Slow?
+
+The first thing I notice is that when I click "deploy" is that I see a bunch of junk on my screen from the `docker_build` call running `pip install`. Didn't we just run `pip install`? My `requirements.txt` didn't change, why do we have to we have to install our dependencies _again_? I thought Docker was supposed to be _good_ at caching!
```
nicks: I appreciate the thorough explanation here. But I'm a little bit worried that this adds a language-specific tutorial on how Docker layers work. Half the users will already know this, and the other half will be confused about whether these optimizations are Tilt-specific.

@ellenkorbes what do you think?

I was thinking that we'd just have a short section like:

```
The first thing we should always do is re-arrange ADDs and RUNs to improve Docker caching. Here's a guide for Python: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

Link to dockerfile

Results
```

and try to avoid having too much language-specific text or images

maiamcc: fair, i streamlined this section--better?

ellenkorbes: Sorry for the late reply. Fair concern, looks good now 👍

PR #472, opened 2020-03-17
guides: python guide doesn't hardcode debug mode

maiamcc:
Hello @nicks, @ellenkorbes,

Please review the following commits I made in branch maiamcc/dont-hardcode-debug:

7db26bf1cb7e98eec819f50eb7099c481b378a73 (2020-03-17 18:10:06 -0400)
guides: python guide doesn't hardcode debug mode

3022e60157e5ca3dc4a106bcd1a3a8810d6755bf (2020-03-17 17:18:45 -0400)
guides: add step to python guide for optimizing dockerfile


docs/example_python.md
```diff
@@ -280,7 +291,14 @@ You can try the server here:
 
 Obviously, this is the simplest possible server we could write; but we hope that this gives you a starting point for running your Flask app (or other Python app) via Tilt!
 
-Other examples:
+## Further Reading
+
+### Other sample Python projects:
+- [abc123](https://github.com/windmilleng/abc123) is a mini microservice app with a Python server called `numbers`
+- The [Live Update Tutorial](http://localhost:4001/live_update_tutorial.html) walks you through optimizing a Python server
```
nicks: :eyes: this tutorial doesn't use python? and also doesn't live on localhost for the reader?

docs/example_python.md
```diff
@@ -225,34 +225,45 @@ Pretty good! But Tilt has some tricks up its sleeve to make it even faster.
...
+        run('touch /app/app.py', trigger='./start-time.txt'),
+
+        # add a congrats message!
+        run('sed -i "s/Hello cats!/{}/g" /app/templates/index.html'.
+            format(congrats)),
 ])
 ```
 
-We've added a new parameter to `docker_build()` with four `live_update` steps that are, in order:
+We've added two new parameters to `docker_build()`. Let's look at the more complex, `live_update`, which contains an array of four steps. When a [live_update](https://docs.tilt.dev/live_update_tutorial.html) is triggered, Tilt will, in order:
```
nicks: the commas around `live_update` are spurious.

Maybe:

```
We've modified two parameters to `docker_build()`: `live_update` and `build_args`.

The `live_update` has 4 steps:
```

docs/example_python.md
```diff
@@ -225,34 +225,45 @@ Pretty good! But Tilt has some tricks up its sleeve to make it even faster.
...
+
+The second additional parameter, `build_args={'flask_env': 'development'}`, corresponds to this Dockerfile change:
+```
+# Default value; will be overridden by build-args, if passed
+ARG flask_env=production
 
-Let's see what this looks like:
+ENV FLASK_ENV $flask_env
+```
+Together, these changes mean that when we build this Dockerfile via this Tiltfile, we set the environment variable `FLASK_ENV=development`, which does a couple of things; most notably for our purposes, it enables hot reloading, so we change `app.py` in place and not have to worry about restarting the server ourselves. (This is a bit of a roundabout way to achieve our goal, but it keeps the app and Dockerfile flexibile enough to be used in production without changes.)
```
nicks: This paragraph feels very long and windy. The subject changes several times. My general impression is that it's pivoting a lot to tackle objections from different directions.

I might shorten it to:

```
The `build_args` configure Flask in development mode.

When in development mode, Flask watches for file changes and reloads the server when necessary.
```

docs/example_python.md
maiamcc: er whoops i meant to link to https://blog.tilt.dev/2019/04/02/fast-kubernetes-development-with-live-update.html

nicks: it might be worth linking to the code repo directly (https://github.com/windmilleng/live_update/tree/master/python)

interestingly, the code repo link in that blog post (https://github.com/windmilleng/live_update/tree/dbentley/initial) is currently broken

PR #475, opened 2020-03-18
document default_registry host_from_cluster param [ch5465]

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/host-from-cluster:

651cd0b51ac2731ff9d1aba3cd73cf28e6ba7947 (2020-03-18 12:37:32 -0400)
document default_registry host_from_cluster param [ch5465]

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -268,7 +268,7 @@ def filter_yaml(yaml: Union[str, List[str], Blob], labels: dict=None, name: str=
     # extract YAMLs of kind "deployment" with metadata.name regex-matching "baz", also matching "bazzoo" and "bar-baz"
     baz_yaml, rest = filter_yaml(rest, name='baz', kind='deployment')
     k8s_yaml(baz_yaml)
-    
+
     # extract YAMLs of kind "deployment" exactly matching metadata.name "foo"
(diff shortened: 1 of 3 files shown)


api/api.py
```diff
@@ -520,7 +520,7 @@ def encode_yaml_stream(objs: List[StructuredDataType]) -> Blob:
   """
   pass
 
-def default_registry(registry: str) -> None:
+def default_registry(registry: str, host_from_cluster: str = None) -> None:
```
nicks: while you're here, i think the first param is called `host`?
https://github.com/windmilleng/tilt/blob/59de82aedb3bad31a65524b5e8d3279eb8650231/internal/tiltfile/docker.go#L519
using a named parameter `registry` won't work

api/api.py
```diff
@@ -529,14 +529,15 @@ def default_registry(registry: str) -> None:
 
   Args:
     registry: The registry that all built images should be renamed to use.
```
nicks: ditto

api/api.py
maiamcc: whoops good catch

PR #477, opened 2020-03-18
api: add os.realpath

jazzdan:


jazzdan:
@nicks for our next trick, do you know how to get nested submodules to work with the docs generator? What I tried in the most recent commit here _doesn't_ work. Any ideas?

maiamcc:
I can maybe help but i'm not sure what you're shooting for--can you explain more?

jazzdan:
@maiamcc Hey, yeah like right now we have `api/modules/os.py` for documenting things in the `os` module. But now that we've added the `os.path` module how do I represent that in this sphinx thing?

`api/modules/os/path.py` didn't work

api/modules/os/path.py
```diff
@@ -0,0 +1,3 @@
+def realpath(path: str) -> str:
```
nicks: wait...do you need to regenerate the html?

jazzdan: That should happen when you run `tilt up`, right?

PR #479, opened 2020-03-19
docs: optimize java guide a bit

nicks:
Hello @maiamcc, @ellenkorbes,

Please review the following commits I made in branch nicks/ch5492/opt:

903fafd1fc06211854c9b061665f0b70c92b9812 (2020-03-19 15:58:32 -0400)
docs: optimize java guide a bit


nicks:
TBH - A few times while writing this, I wondered if we should just tell people to use Jib, and provide better documentation on how to use live_update with Jib.

It really is an amazing tool. But it also deliberately hides a lot of details about Docker. Jib makes it very hard to debug what it's doing when things go wrong (e.g., you can't shell into a Jib container by default: https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#where-is-bash). So I'm a little bit nervous about recommending it.

nicks:
Ya, I eventually got Jib working with live_update, and it was AWFUL https://github.com/windmilleng/tilt-example-java/tree/master/102-jib-live-update

docs/example_java.md
```diff
@@ -223,47 +223,129 @@ Let's see what this looks like!
...
-Here's our [new Tiltfile](https://github.com/windmilleng/tilt-example-java/blob/master/3-recommended/Tiltfile) 
+Here's our [new Tiltfile](https://github.com/windmilleng/tilt-example-java/blob/master/4-recommended/Tiltfile) 
 with the following new code:
 
 ```python
+local_resource(
+  'example-java-compile',
+  './gradlew bootJar && ' +
+  'unzip -o build/libs/example-0.0.1-SNAPSHOT.jar -d build/jar-staging && ' +
+  'rsync --inplace --checksum -r build/jar-staging/ build/jar',
```
ellenkorbes: Clever! 👏

(Though I wonder if the added indirection of needing a staging folder won't make people confused.)

ellenkorbes:
Why was Jib + live_update awful?

docs/example_java.md
nicks: ya, i wonder if live_update should have some sort of checksum feature, so that you don't need to do this.

nicks:
re: why Jib + live_update is awful
- Tilt is really NOT built for the use-case where you have one local command to build the image, and an entirely different local command for generating the file to sync. The example I drew up is VERY subtle and non-obvious (there's a local resource that builds the classfiles, and a custom_build command that runs whenever classfiles are generated)
- Jib is hyper-focused on optimized images, not developer-friendly images. If you want to add things to the image (bash, entr), the flow is awkward
- Jib doesn't behave well with local-only images (images that don't live in a remote registry), tho I honestly think this is a bug we should get them to fix.

Want to file additional tickets about this just so I remember it all while it's in my head, but not totally sure what the right answers are

PR #483, opened 2020-03-23
nodejs onboarding guide [ch4827]

maiamcc:
example code here: https://github.com/windmilleng/tilt-example-nodejs/

screenshots/snapshots pending until all the kinks in the app have been worked out.


docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+app.get('/', function(req, res) {
+    res.sendFile(path.join(__dirname + '/index.html'));
+});
 
-[Servantes](https://github.com/windmilleng/servantes) is a personalized homepage
-app implemented with many microservices in multiple languages. You can run it
-with the commands
+app.listen(8000, () => {
+    console.log('Server running at http://localhost:8000/');
+});
```
ellenkorbes: I find it a bit confusing that in both `listen()` and `get()` we're passing a function as the second argument, but one uses the regular function syntax, and the other uses the arrow function syntax. While technically correct, to me it feels like someone using Shakespearean English and Brooklyn English in the same sentence.

I'm not a JS developer though, so maybe I'm missing something.

docs/example_python.md
```diff
@@ -232,11 +232,11 @@ docker_build('example-python-image', '.', build_args={'flask_env': 'development'
 ])
 ```
 
-We've added two new parameters to `docker_build()`: `build args` amd `live_update`. Let's look at the latter first.
+We've added two new parameters to `docker_build()`: `build_args` amd `live_update`. Let's look at the latter first.
```
ellenkorbes: `s/amd/and`?

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+
+Let's fix this.
+
+In this example, we're going to take you through a simple “hello world” server written in NodeJS that uses [Mustache](https://mustache.github.io/) for templating and [Express](https://expressjs.com/) for serving.
+
+We'll use Tilt to:
+
+- Run the server on Kubernetes
+- Measure the time from a code change to a new process
+- Optimize that time for fast feedback
```
ellenkorbes: `s/fast/faster`?

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+
+Obviously, this particular server doesn't do much, but it's still a useful example to confirm that Tilt is working
```
ellenkorbes:
```suggestion
This particular example doesn't do much, but it is useful to confirm that Tilt is working as expected in your environment.
```

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
 
-[Servantes](https://github.com/windmilleng/servantes) is a personalized homepage
-app implemented with many microservices in multiple languages. You can run it
-with the commands
+app.listen(8000, () => {
+    console.log('Server running at http://localhost:8000/');
+});
+```
+
+To start this server on Kubernetes, we need 3 configs:
```
ellenkorbes:
```suggestion
To start this server on Kubernetes, we need 3 config files:
```

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+3. And finally, a [Tiltfile](https://github.com/windmilleng/tilt-example-nodejs/blob/master/0-base/Tiltfile) that ties them together:
+
+```python
+docker_build('example-nodejs-image', '.')
+k8s_yaml('kubernetes.yaml')
+k8s_resource('example-nodejs', port_forwards=8000)
+```
+
+The first line tells Tilt to build an image with the name `example-nodejs-image`
+in the directory `.` (the current directory).
```
ellenkorbes:
```suggestion
in the current directory.
```

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+
+The second line tells Tilt to load the Kubernetes
+[Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment)
+YAML. The image name in the `docker_build` call must match the container `image`
+reference in the `example-nodejs` Deployment.
+
+The last line configures port-forwarding so that your server is
+reachable at http://localhost:8000/. The resource name in the `k8s_resource` call
```
ellenkorbes:
```suggestion
reachable at `localhost:8000`. The resource name in the `k8s_resource` call
```

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+
+Before we try to make this faster, let's measure it.
+
+In addition to running things in your cluster, Tilt can run commands locally.
+You can direct Tilt to execute existing scripts or arbitrary shell commands on your own machine.
+
+We want to measure the time from code change to new process. To do that, we'll
+use [`local_resource`](local_resource.html), which lets you locally
+run scripts, shell code, or servers, and manage them from your sidebar like any
+other Tilt resource.
```
ellenkorbes:
```suggestion
Using [`local_resource`](local_resource.html), you can direct Tilt to execute existing scripts or arbitrary shell commands on your own machine, and manage them from your sidebar like any other Tilt resource.
```

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+
+| Approach | Deploy Time (after initial)
+|---|---|
+| Naive | 11.31-14.21s |
+| Optimized Dockerfile | 3.25-4.12s |
+{:.benchmark-report}
+
+Pretty good! But Tilt has some tricks up its sleeve to make it even faster.
+
+## Step 3: Let's Optimize It EVEN MORE
```
ellenkorbes:
```suggestion
## Step 3: Let's Optimize It _Even More_
```

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+2. the `build_args` parameter corresponds to this Dockerfile change:
+    ```
+    # Default value; will be overridden by build_args, if passed
+    ARG node_env=production
+
+    ENV NODE_ENV $node_env
+    ```
+   Together, these changes mean that when we build this Dockerfile via this Tiltfile, we set the env var `$NODE_ENV=development`, and our `yarn install` call will install dev dependencies.
+
+With the help of Tilt, we've done a pretty cool thing: we've set it up so this Dockerfile can run the app for development (when invoked via Tilt) or for production (when invoked otherwise)---because we at Tilt think that the closer your dev and prod configs are, the more unpleasant surprises you avoid.
```
ellenkorbes:
```suggestion
With the help of Tilt, we've done a pretty cool thing: we've set it up so this Dockerfile can run the app for development (when invoked via Tilt) or for production (when invoked otherwise).
```

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+
+The other new addition to our Tiltfile is the `live_update` arg to `docker_build`, which enables super-fast in-place updates of your app. When a [live_update](https://docs.tilt.dev/live_update_tutorial.html) is triggered, Tilt will, in order:
```
ellenkorbes: `s/arg/argument`

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+Let's see what this new configuration looks like in action:
+
+<figure>
+  <a class="is-image" title="Tilt state after a live_update" href="https://cloud.tilt.dev/snapshot/todo">
+    <img src="assets/docimg/example-nodejs-liveupdate.png">
+  </a>
+  <figcaption>The result of a live_update. (Click the screenshot to see an interactive view.)</figcaption>
+</figure>
+
+Tilt was able to update the container in less than two seconds! (And a chunk of that time was overhead from nodemon, not from Tilt.)
```
ellenkorbes:
```suggestion
Tilt was able to update the container in less than two seconds! (That's roughly half from Tilt's `live_reload`, and half from nodemon's overhead.)
```
Felt a bit too defensive before.

docs/example_nodejs.md
```diff
@@ -1,34 +1,313 @@
...
+| Naive | 11.31-14.21s |
+| Optimized Dockerfile | 3.25-4.12s |
+| With live_update | 1.1-1.8s |
+{:.benchmark-report}
+
+You can try the server here:
+
+[Recommended Structure](https://github.com/windmilleng/tilt-example-nodejs/blob/master/3-recommended)
+
+Obviously, this is the simplest possible server we could write; but we hope that this gives you a starting point for running your NodeJS app via Tilt!
```
ellenkorbes:
```suggestion
This is a very simple example, but we hope it gives you a good starting point for running your NodeJS app via Tilt!
```

docs/example_nodejs.md
maiamcc: oh, there's no good reason from this, I was just stealing them from diff guides. will fix.

docs/example_nodejs.md
maiamcc: FINE I'll use less words thhbt. (Good call.)

docs/example_nodejs.md
maiamcc: ha fair. changed it to:
> Tilt and nodemon together updated the container in less than two seconds!
wdyt?

maiamcc:
Accepted all your suggestions specific to this doc, left a handful that are for shared copy and will apply those in a separate PR.

maiamcc:
@nicks or @jazzdan can i get some JS engineer eyes?

docs/example_nodejs.md
```diff
@@ -1,34 +1,314 @@
...
+
+All the code is in this repo:
+
+[tilt-example-nodejs](https://github.com/windmilleng/tilt-example-nodejs)
+
+To skip straight to the fully optimized setup, go to this subdirectory:
+
+[Recommended Tiltfile](https://github.com/windmilleng/tilt-example-nodejs/tree/master/3-recommended)
+
+_Quick note: Tilt's first deployment of a service takes a few seconds longer than subsequent ones, due to some behind-the-scenes setup. Measurements in this guide focus on non-initial builds._
```
nicks: A first-time Tilt user isn't going to understand or even care about this comment, let's either move this to a footnote or axe it entirely

docs/example_nodejs.md
```diff
@@ -1,34 +1,314 @@
...
+});
+
+function getSecsSinceDeploy() {
+    let curTimeMs = new Date().getTime();
+    let contents = fs.readFileSync('/app/start-time.txt', 'utf8');
+    let startTimeMs  = parseInt(contents.trim()) / 10**6;
+    return ((curTimeMs - startTimeMs) / 10**3).toFixed(2)
+}
+```
+Whenever the app starts up, it calls `getSecsSinceDeploy()`, does the math to figure out the time elapsed since the timestamp in `start-time.txt`, and stores that value in a global variable; the app then templates that value into `index.mustache` so that it shows up in the served webpage. (We've added a new dep for this purpose: the templating engine [mustache-express](https://www.npmjs.com/package/mustache-express).)
```
nicks: this is a very complicated paragraph, can we simplify it a bit?

docs/example_nodejs.md
```diff
@@ -1,34 +1,314 @@
...
+Here's what it looks like when we build with our new Dockerfile:
+
+<figure>
+  <a class="is-image" title="Result of clicking the button  on the “deploy” resource" href="https://cloud.tilt.dev/snapshot/todo">
+    <img src="assets/docimg/example-nodejs-optimized-dockerfile.png">
+  </a>
+  <figcaption>"RUN pip install..." now uses the cache instead of actually running a long, slow command. (Click the screenshot to see an interactive view.)</figcaption>
+</figure>
+
+Hooray, we're now using the cache instead of running `yarn install` for every single build. (For more on the principles at work here, [check out this guide](http://bitjudo.com/blog/2014/03/13/building-efficient-dockerfiles-node-dot-js/).)
```
nicks: I would probably point to something an actively-maintained guide like https://nodejs.org/en/docs/guides/nodejs-docker-webapp/, rather than a blog post. The blog post will go out of date. We don't expect the author to go back and fix it. Whereas a widely-used guide gets more traffic and fixes.

docs/example_nodejs.md
```diff
@@ -1,34 +1,314 @@
...
+        # reloads so that it will reflect the new startup time
+        run('touch /app/index.js', trigger='./start-time.txt'),
+
+        # add a congrats message!
+        run('sed -i "s/Hello cats!/{}/g" /app/views/index.mustache'.
+            format(congrats)),
+])
+```
+We've added some new parameters to `docker_build` that tell the container to use nodemon:
+1. the `entrypoint` parameter overrides the `ENTRYPOINT` specified in the Dockerfile; now when the container executes, it will run `yarn run nodemon /app/index.js`
```
nicks: These should be capitalized, like the other lists

docs/example_nodejs.md
```diff
@@ -1,34 +1,314 @@
...
+2. the `build_args` parameter corresponds to this Dockerfile change:
+    ```
+    # Default value; will be overridden by build_args, if passed
+    ARG node_env=production
+
+    ENV NODE_ENV $node_env
+    ```
+   Together, these changes mean that when we build this Dockerfile via this Tiltfile, we set the env var `$NODE_ENV=development`, and our `yarn install` call will install dev dependencies.
+
+With the help of Tilt, we've done a pretty cool thing: we've set it up so this Dockerfile can run the app for development (when invoked via Tilt) or for production (when invoked otherwise).
```
nicks: I'm not sure this paragraph is necessary. If I'm just trying to set up my app, I don't think this would be an important or notable thing I would care about.

docs/example_nodejs.md
maiamcc: good clal

docs/example_nodejs.md
maiamcc: Sure, but if we're trying to teach _best practices_ then I think this IS a thing we want to highlight -- we're pretty opinionated that if possible, you should use t same dockerfile for prod and dev, and just use Tilt to change the dev-specific bits, right?

docs/example_nodejs.md
maiamcc: that's a fair point but the blog post is really precisely about the concept I'm alluding to here, whereas you need to dig for it in the nodejs guide. (also the nodejs guide links to the same blogpost...)

maiamcc:
Added images (with some templating logic to make it easier to update them, keep logic uniform, etc.) -- possibly overkill but I had a weird morning and this was a nice distraction.

docs/example_nodejs.md
nicks: I don't want to make a bigger deal with it than this particular edit merits. 

But your comment here gets a couple things conflated/confused that we should talk about more in detail. Because they're good questions! And worth talking about!

1) The terms "opinionated tools" and "best practices" tend to get thrown around a lot in the tech community, without much introspection about what, exactly, it's worth being opinionated about and what assumptions those opinions are based on.

2) I don't think the opinion "you should use the same Dockerfile for dev and prod AT ALL COSTS" is a particularly good opinion or best practice

3) If we were going to settle on a best practice, it would be something like "prefer layers over branches" NOT "use the same dockerfile at all costs." And I'd recommend we add docker_build(env) and remove the build_args here. That tends to be more what I see in the wild.

4) The approach this example takes ("use build_args to parameterize for dev or prod") is a very narrow instance. The instant jump from specific example to broad principle comes off as very Hacker News comment-y in tone (i.e., "this worked for my codebase, so it should work for all codebases!"). 

5) There's a big gap between using a best practice in the example, and trying to explain that best practice effectively. I think trying to explain a best practice in an off-hand comment mid-guide isn't going to convey what you want.

Sorry, I know that's a lot to digest, but I hope it's helpful!

docs/example_nodejs.md
nicks: thanks, this is a big improvement!

docs/example_nodejs.md
maiamcc: hmmmm okay that makes sense, and i appreciate that framing. I'll cut it.

maiamcc: hmmmm okay that makes sense, and i appreciate that framing. I'll cut it.

PR #488, opened 2020-03-25
guides: change all guides to use includes for figures and captions

maiamcc:
This applies new functionality from #484 to all guides.


maiamcc:
FYI @nicks you had the wrong snapshot linked for Golang naive--I made one, but if you particularly care about the snapshots being homogenous (i.e. having your name in the directory paths) you can go ahead and make your own and send me the link to insert.

nicks:
wait what? where did this example_guide_image.html pattern come from? My apologies. I missed this. 

In the future, when we're doing reviews, I expect that the changes you're making are only responding to comments. I wasn't expecting (or watching out for) experimental new template architectures!

Have you considered using template arguments? https://jekyllrb.com/docs/includes/ They're way more intuitive and don't require a global map

nicks:
oh, and re: "you had the wrong snapshot linked for Golang naive" - I'm not sure I see what you're referring to. there are no new Go pngs in this PR? Can you send this fix in a separate PR?

nicks:
(and also i really like the idea of using an include template for this, just want to make sure we use it effectively!)

maiamcc:
> oh, and re: "you had the wrong snapshot linked for Golang naive" - I'm not sure I see what you're referring to. there are no new Go pngs in this PR? Can you send this fix in a separate PR?

sure, done in #491

maiamcc:
> Have you considered using template arguments? https://jekyllrb.com/docs/includes/ They're way more intuitive and don't require a global map

i did look at template arguments, and decided on the global map because a. it makes it easier to do defaults where the copy is the same, so we can better centralize copy, and b. I like the fact that if I say, want to replace the golang base snapshot, it's really easy to find it and replace it, and I don't need to scroll through a bunch of nearly-identical invocations to find the one I care about and replace it. (The big global map is also useful if you want to say, replace all the snapshots at once.)

That was my reasoning but also I don't feel super strongly about it, if you feel more strongly than I do, I'm happy to use args!

nicks:
Ya, I think centralizing copy is a good goal. Centralizing it in a global map in a figure include that reads global variable isn't a great abstraction to accomplish that goal, and I'd prefer duplication over a bad abstraction. Happy to discuss or propose alternative approaches.

This is a great blog post on this topic that you might have seen before: https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction

maiamcc:
Mmmm fair -- and interesting article! Will try to keep that in mind.

maiamcc:
fyi @jazzdan i fixed up the C# guide to match the new templating stuff!

PR #489, opened 2020-03-25
guides: styling to relate own-line links to the line above [ch5388]

maiamcc:
@hyu i didn't get to the accompaning github icon suggestion, but
this seemed like a good start!

### DIFF
--- a/docs/example_go.md
+++ b/docs/example_go.md
@@ -20,11 +20,11 @@ We'll use Tilt to:
 
 All the code is in this repo:
 
-[tilt-example-go](https://github.com/windmilleng/tilt-example-go)
+[tilt-example-go](https://github.com/windmilleng/tilt-example-go){:.attached-above}
 
 To skip straight to the fully optimized setup, go to this subdirectory:
 
-[Recommended Tiltfile](https://github.com/windmilleng/tilt-example-go/blob/master/3-recommended/Tiltfile)
+[Recommended Tiltfile](https://github.com/windmilleng/tilt-example-go/blob/master/3-recommended/Tiltfile){:.attached-above}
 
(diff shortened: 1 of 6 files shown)


maiamcc:
Screen Shot 2020-03-25 at 6 31 48 PM

(Is the bold too much?)

docs/example_go.md
```diff
@@ -18,13 +18,11 @@ We'll use Tilt to:
 - Measure the time from a code change to a new process
 - Optimize that time for fast feedback
 
-All the code is in this repo:
+All the code is in this repo: [tilt-example-go](https://github.com/windmilleng/tilt-example-go)
+{:.link-own-line}
```
nicks: Putting this in one paragraph, and then using css to move it into multiple paragraphs, seems very semantically odd.

Have you considered doing this with pure CSS and not making any text changes at all?

e.g.,

```
p > a:only-child {
  text-indent: 2em;
  font-weight: bold;
}
```

maiamcc: I considered that but the main thing I'm trying to do is get rid of the extra whitespace between the link and the paragraph above, so it's clear what bit of text the link "belongs" to; and messing with negative margins seemed foolhardy if say, the paragraph margins change from desktop to mobile. if the margins are consistent enough though, I'd happily do
```
p > a:only-child {
  text-indent: 2em;
  font-weight: bold;
  margin-top: -1em;
}
```
or whatever.

nicks: ya, i like that suggestion. we typically set margin-bottom to $spacing-unit / 2, so maybe do margin-top: - $spacing-unit / 4

maiamcc: cool, done.

PR #508, opened 2020-04-07
docs: clarify docs for ignore and only params

maiamcc:
a user in #tilt expressed confusion with the docs on `ignore` and `only`
(e.g. they didn't know what directory they're evaluated relative to).
This PR attempts to clarify the API docs and the File Changes Guide
on this point.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -93,7 +93,7 @@ def docker_build(ref: str, context: str, build_args: Dict[str, str] = {}, docker
 
     docker build /path/to/code -t myregistry/myproj/backend
 
-  For more information on the `ignore` and `only` parameters, see our `Guide to File Changes <http://localhost:4001/file_changes.html>`_.
+  For more information on the `ignore` and `only` parameters, see our `Guide to File Changes </file_changes.html>`_.
 
   Note that you can't set both the `dockerfile` and `dockerfile_contents` arguments (will throw an error).
 
--- a/api/api.py
+++ b/api/api.py
@@ -36,7 +36,7 @@ def set_team(team_id: str) -> None:
 
   Sends usage information to Tilt Cloud periodically.
   """
-  pass  
+  pass
 
 def sync(local_path: str, remote_path: str) -> LiveUpdateStep:
   """Specify that any changes to `localPath` should be synced to `remotePath`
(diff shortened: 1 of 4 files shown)


api/api.py
```diff
@@ -97,8 +107,8 @@ def docker_build(ref: str, context: str, build_args: Dict[str, str] = {}, docker
     dockerfile_contents: raw contents of the Dockerfile to use for this build.
     live_update: set of steps for updating a running container (see `Live Update documentation <live_update_reference.html>`_).
     match_in_env_vars: specifies that k8s objects can reference this image in their environment variables, and Tilt will handle those variables the same as it usually handles a k8s container spec's ``image`` s.
-    ignore: set of file patterns that will be ignored. Ignored files will not trigger builds and will not be included in images. Follows the `dockerignore syntax <https://docs.docker.com/engine/reference/builder/#dockerignore-file>`_.
-    only: set of file paths that should be considered for the build. All other changes will not trigger a build and will not be included in images. Inverse of ignore parameter. Only accepts real paths, not file globs.
+    ignore: set of file patterns that will be ignored. Ignored files will not trigger builds and will not be included in images. Follows the `dockerignore syntax <https://docs.docker.com/engine/reference/builder/#dockerignore-file>`_. Patterns will be evaluated relative to ``context``.
```
landism, line 110: maybe s/``context``/the build's ``context``/ here and below? I feel like it'd be easy to wonder if this were referring to some kind of Tiltfile context.

docs/custom_build.md
```diff
@@ -123,6 +123,22 @@ Tilt's `docker_build` supports other options. The most impactful is [Live Update
 
 `custom_build` supports most other options of `docker_build`, and a few specific to non-Docker container builders. If you find an option you think should exist but doesn't, let us know in the `#tilt` channel in [Kubernetes Slack](http://slack.k8s.io).
 
+### Adjust Filewatching wtih `ignore`
```
landism, line 126:
```suggestion
### Adjust Filewatching with `ignore`
```

maiamcc:
> mind doing `local_resource`, too?
> 
> unfortunately it's not consistent with `custom_build`, and ignores are relative to the Tiltfile dir

le sigh. yes, will do!

api/api.py
maiamcc: good call, changed to "relative to the `context` parameter", better?

maiamcc:
> mind doing `local_resource`, too?
> 
> unfortunately it's not consistent with `custom_build`, and ignores are relative to the Tiltfile dir

...wait this is stupid, shouldn't we just make them the same?

landism:
> ...wait this is stupid, shouldn't we just make them the same?

https://app.clubhouse.io/windmill/story/6088/reconcile-custom-build-and-local-resource-ignore-wd

maiamcc:
Cool, specifying the current local_resource behavior with a mighty hope that we fix it soon 😅

PR #509, opened 2020-04-08
api: Fix typo in code example of filter_yaml

victorwuky:


victorwuky:
@nicks : Can you please double check?

PR #511, opened 2020-04-09
Update tutorial to emphasize web UI and not TUI

victorwuky:


victorwuky:
@nicks : Can you please review. Thanks.

PR #516, opened 2020-04-15
docs: document version_settings' new constraint param

landism:
documents https://github.com/windmilleng/tilt/pull/3172

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -776,7 +776,12 @@ def version_settings(check_updates: bool = True, constraint: str = "") -> None:
                    available.
     constraint: If non-empty, Tilt will check its currently running version against
                 this constraint and generate an error if it doesn't match.
-                See the `constraint syntax documentation <https://github.com/blang/semver#ranges>`_.
+                Examples:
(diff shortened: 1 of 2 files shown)


api/api.py
```diff
@@ -767,12 +767,16 @@ def analytics_settings(enable: bool) -> None:
...
-  By default this is set to True.
+def version_settings(check_updates: bool = True, constraint: str = "") -> None:
+  """Controls Tilt's behavior with regard to its own version.
 
   Args:
-    check_updates: whether or not to check for new versions of Tilt on GitHub.
+    check_updates: If true, Tilt will check GitHub for new versions of itself
+                   and display a notification in the web UI when an upgrade is
+                   available.
+    constraint: If non-empty, Tilt will check its currently running version against
```
maiamcc, line 777: not strictly necessary because you link to the syntax right there, but an example or two in this doc would make things a bit more accessible/one-stop-shop.

landism: good call, thanks

PR #528, opened 2020-04-20
Specify local API function runs on host machine

victorwuky:


api/api.py
```diff
@@ -345,7 +345,7 @@ def load(path: str, *args):
   """
 
 def local(cmd: str, quiet: bool = False) -> Blob:
-  """Runs cmd, waits for it to finish, and returns its stdout as a ``Blob``
+  """Runs cmd on the *host* machine, waits for it to finish, and returns its stdout as a ``Blob``
```
nicks, line 348: i don't think this is anything we've heard people be confused about, can you elaborate a bit on why you think it's important to accentuate?

victorwuky: I was confused (https://windmill-eng.slack.com/archives/CBG86T5U4/p1587396724052800?thread_ts=1587394846.052000&cid=CBG86T5U4).

The docs of `local_resource` use this language (right below it: https://docs.tilt.dev/api.html#api.local). And so when I read it, I was confused because if both functions work locally, I'd expect the docs to say so. The fact that the docs doesn't mention it, or that the explanations in each function don't mention each other, confused me as a user.

nicks: fair! though you need to run `make api` to regenerate the html files

PR #535, opened 2020-04-21
docs: add more detail on ignore/only for custom_build/local_resource

nicks:
Hello @landism, @jazzdan,

Please review the following commits I made in branch nicks/custombuild:

a2b8d2b3b876865e80c93d0857c1acb3ba0c3c91 (2020-04-21 13:15:48 -0400)
docs: add more detail on ignore/only for custom_build/local_resource

### DIFF
--- a/docs/file_changes.md
+++ b/docs/file_changes.md
@@ -50,6 +50,14 @@ a directory to build. Tilt will watch the entire directory.
 Whenever a file in that directory changes, Tilt will re-build the image,
 then deploy any Kubernetes resources that depend on that image.
 
+### Custom and Local Resource Builds
+
+When you build other types of resources (like `custom_build()` and
+`local_resource()`), you can specify a file system path or a list of paths as
+`deps`.
+
+Whenever a file under that path changes, Tilt will re-run the specified scripts.
+
 ## Ignoring Files
 
 ### .git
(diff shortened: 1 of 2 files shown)


nicks:
(people ask about this in the tilt channel occasionally)

docs/file_changes.md
```diff
@@ -130,7 +145,21 @@ is equivalent to having a `.dockerignore` file that looks like:
...
+
+Other build types don't have an `only=` parameter.
+
+In a `docker_build`, you're creating a docker build context to copy into the
+image.  The Docker build context is [an important
+concept](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#understand-build-context)
+for writing efficient Dockerfiles. The `only=` parameter lets you select
+specific directories in the build context.
+
+In other build types, you can specify multiply directories with `deps=`.
```
landism, line 162: I think we could get a bit more explicit about the difference, along the lines of:

For `docker_build`, `only` filters what files are actually available to the docker build within the build context. Other types of builds are simply able to access all files, and `deps=` is used to specify which paths trigger them.

Or, if you don't like that framing, in this last line: talking about "multiple" seems like it'd confuse more than clarify, and I'd swap out "directories" for "paths", since they can be files.

docs/file_changes.md
```diff
@@ -50,6 +50,15 @@ a directory to build. Tilt will watch the entire directory.
 Whenever a file in that directory changes, Tilt will re-build the image,
 then deploy any Kubernetes resources that depend on that image.
 
+### Custom and Local Resource Builds
+
+When you build other types of resources (like `custom_build()` and
+`local_resource()`), you can specify a directory or a set of directories
```
landism, line 56: here and below: deps can be files

nicks: i changed it to 'paths', which i think most people will understand to be files or directories

docs/file_changes.md
nicks: i like that wording! done

PR #542, opened 2020-04-24
docs: document config.define_string and define_bool

landism:
whoops never got these into docs

### DIFF
--- a/api/modules/config.py
+++ b/api/modules/config.py
@@ -27,7 +27,7 @@ def define_string(name: str, args: bool=False, usage: str="") -> None:
     Allows the user invoking Tilt to configure a key named ``name`` to be in the
     dict returned by :meth:`parse`.
 
-    To set a flag of this type named, e.g., `foo`, at runtime to "bar", run ``tilt up -- --foo bar``.
+    For instance, at runtime, to set a flag of this type named `foo` to value "bar", run ``tilt up -- --foo bar``.
 
     See the `Tiltfile config documentation <tiltfile_config.html>`_ for examples
     and more information.
@@ -49,7 +49,13 @@ def define_bool(name: str, args: bool=False, usage: str="") -> None:
     Allows the user invoking Tilt to configure a key named ``name`` to be in the
     dict returned by :meth:`parse`.
 
-    To set a flag of this type named, e.g., `foo`, to True, run ``tilt up -- --foo``.
+    For instance, at runtime, to set a flag of this type named `foo` to value `True`, run ``tilt up -- --foo``.
+    To set a value to ``False``, you can run ``tilt up -- --foo=False``, or use a default value, e.g.:
+    ```python
+    config.define_bool('foo')
+    cfg = config.parse()
+    do_stuff = cfg.get('foo', False)
+    ```
 
     See the `Tiltfile config documentation <tiltfile_config.html>`_ for examples
     and more information.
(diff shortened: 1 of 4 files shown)


api/modules/config.py
```diff
@@ -20,6 +20,51 @@ def define_string_list(name: str, args: bool=False, usage: str="") -> None:
...
     """
 
+def define_string(name: str, args: bool=False, usage: str="") -> None:
+    """
+    Defines a config setting of type `str`.
+
+    Allows the user invoking Tilt to configure a key named ``name`` to be in the
+    dict returned by :meth:`parse`.
+
+    To set a flag of this type named, e.g., `foo`, at runtime to "bar", run ``tilt up -- --foo bar``.
```
maiamcc, line 30: Reorder this clause for clarity
```suggestion
    For instance, at runtime, to set a flag of this type named `foo` to value "bar", run ``tilt up -- --foo bar``.
```

api/modules/config.py
```diff
@@ -20,6 +20,51 @@ def define_string_list(name: str, args: bool=False, usage: str="") -> None:
...
+    """
+
+def define_bool(name: str, args: bool=False, usage: str="") -> None:
+    """
+    Defines a config setting of type `bool`.
+
+    Allows the user invoking Tilt to configure a key named ``name`` to be in the
+    dict returned by :meth:`parse`.
+
+    To set a flag of this type named, e.g., `foo`, to True, run ``tilt up -- --foo``.
```
maiamcc, line 52: see above, reorder this clause so "e.g." comes first

maiamcc: also maybe worth giving an example of how to set to false cuz that's a little less intuitive

api/modules/config.py
```diff
@@ -20,6 +20,51 @@ def define_string_list(name: str, args: bool=False, usage: str="") -> None:
...
+    dict returned by :meth:`parse`.
+
+    To set a flag of this type named, e.g., `foo`, at runtime to "bar", run ``tilt up -- --foo bar``.
+
+    See the `Tiltfile config documentation <tiltfile_config.html>`_ for examples
+    and more information.
+
+    Args:
+      name: The name of the config setting
+      args: If False, the config setting is specified by its name. (e.g., if it's named "foo",
```
maiamcc, line 37: sigh.

docs/tiltfile_config.md
```diff
@@ -146,12 +146,25 @@ cfg = config.parse()
 config.set_enabled_resources(cfg.get('args', []))
 ```
 
+## Flag types
+Here are some sample usages of different types of flags that are supported:
+```python
+# list of string
```
maiamcc, line 152: if you keep this format, def put in "paragraph breaks" in your code block (e.g. line break before "single string".

consider
```
# single string
# invocation: tilt up -- ...
```
to make it clear what that line of comment is for (took me a second)

docs/tiltfile_config.md
```diff
@@ -146,12 +146,25 @@ cfg = config.parse()
...
+config.define_bool("local")
+```
+
 ## Future Directions
 This section describes places we expect the config to go. Let us know if any of
 these would be particularly helpful to you and your team.
 ### More Kinds of Settings
-You should be able to define more kinds of settings. For example, a bool, an
-an enum, or a singly-valued string. By moving error-checking to the built-in
+You should be able to define more kinds of settings. For example, an enum. By moving error-checking to the built-in
```
maiamcc, line 167:
```suggestion
You should be able to define more kinds of settings (for example, an enum). By moving error-checking to the built-in
```

docs/tiltfile_config.md
maiamcc: ##### Single String
###### Invocation: `tilt up -- ...`
```
str(whatever)
```

PR #547, opened 2020-04-28
update docs to reference restart_process extension instead of restart_container live update step when possible [ch5945]

maiamcc:
this is necessarily kind of ugly b/c this extension should only be
used SOMETIMES :-/

this seemed like an okay first pass? next i'll revise the live_update tutorial


maiamcc:
Holding off on these docs until we've had the extension beta-tested

maiamcc:
This got fixed elsewhere, closing.

PR #560, opened 2020-05-04
Specify snapshot data in FAQ entry about disabling snapshots

victorwuky:
Based on https://app.clubhouse.io/windmill/story/6731/when-disable-snapshots-is-set-usage-is-not-captured-in-tilt-cloud-team-page#activity-6782

### DIFF
--- a/docs/snapshots.md
+++ b/docs/snapshots.md
@@ -49,6 +49,10 @@ A snapshot is accessible to anyone who has the link, which includes an unguessab
 Providing stricter access control is on our roadmap. If this is important to you,
 [let us know](https://tilt.dev/contact)!
 
+### Q: My company doesn't allow sharing snapshot data with a SaaS. How can I disable snapshots?
+Add the [`disable_snapshots`](https://docs.tilt.dev/api.html#api.disable_snapshots)
+directive to your Tiltfile.
+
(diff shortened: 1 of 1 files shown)


nicks:
hmmm...this still seems valid to me. I didn't totally understand from your comment why you want to remove it, can you elaborate?

victorwuky:
> hmmm...this still seems valid to me. I didn't totally understand from your comment why you want to remove it, can you elaborate?

@nicks : I updated it again. Instead of removing it, I specified it is about snapshot data only. As we build out Tilt Cloud, we are "sending data" in a variety of ways and different types of data. You could interpret adding a team name and changing it as "sending data". Since I think we want this feature of disabling snapshots to be only about snapshots, I think this makes it clear and doesn't mislead people into thinking about the situation where they want to disable Tilt Cloud overall. (Even if I'm over-thinking it, this I think is still a good change and a good clarification.)

nicks:
LGTM

though I might focus it a bit more on the concern we've actually heard - i.e., "I'm worried about app devs accidentally posting snapshots with secrets or private data." because then we can also plug the private snapshots ticket

victorwuky:
@nicks : Do you mind reviewing one more time?

PR #561, opened 2020-05-04
docs: promote team version

landism:
image


docs/teams.md
```diff
@@ -7,30 +7,43 @@ layout: docs
...
+<figure>
+    <img src="/assets/img/team-promote-version.png" class="no-shadow" alt="Promote Team Version">
+</figure>
 
 ## Add users to your team
 
 New users on the team page will appear with the role `None` as they run Tilt. A team owner (`Owner`) can change any other user's role to `Member` or `Owner`.
-An `Owner` cannot change their own role. Only `Owner`s can change the team name. Both `Member`s and `Owners` can view the team page.
\ No newline at end of file
+An `Owner` cannot change their own role. Only `Owner`s can change the team name. Both `Member`s and `Owners` can view the team page.
```
victorwuky, line 49:
```suggestion
An `Owner` cannot change their own role. Only `Owner`s can change the team name. Both `Member`s and `Owner`s can view the team page.
```

docs/teams.md
```diff
@@ -7,30 +7,43 @@ layout: docs
...
-You'll now see all folks who run Tilt with the updated Tilfile, appear on the team page, including when they last ran Tilt and the version they are using. Ask users to sign in to Tilt Cloud in Tilt for their name to appear on the team page. Otherwise, they will appear as anonymous.
+You'll now see all folks who run Tilt with the updated Tiltfile, appear on the team page, including when they last ran Tilt and the version they are using. Ask users to sign in to Tilt Cloud in Tilt for their name to appear on the team page. Otherwise, they will appear as anonymous.
 
 <figure>
     <img src="/assets/img/last-seen-versions.png" class="no-shadow" alt="Last seen versions">
 </figure>
 
+## Vet Tilt releases before your team upgrades
+
+Tilt's web UI will nudge users to upgrade when a new release is available.
```
victorwuky, line 34:
```suggestion
Tilt's web UI will nudge users to upgrade when a new release is available, at the bottom right of the screen.
```

docs/teams.md
```diff
@@ -7,30 +7,43 @@ layout: docs
...
+
+When running Tilt with a Tiltfile that specifies a team id, the default behavior is that users are only notified about Tilt releases once they're one week old.
```
victorwuky, line 36:
```suggestion
When running Tilt with a Tiltfile that specifies a team id, non-`Owner` users are only notified about Tilt releases once they're one week old.
```

docs/teams.md
```diff
@@ -7,30 +7,43 @@ layout: docs
...
+
+However, the team's *owners* are notified of Tilt releases as soon as they come out, giving them an opportunity to vet the new version. If no action is taken, Tilt will begin notifying other team members of a release once it's a week old.
```
victorwuky, line 38:
```suggestion
However, the team's *owners* are notified of Tilt releases as soon as they come out, giving them an opportunity to vet the new version. If no action is taken, Tilt will begin notifying other non-`Owner`s of a release once it's a week old.
```

Let's be specific in who these users are, and in particular, avoid "members" since that's a role.

docs/teams.md
```diff
@@ -7,30 +7,43 @@ layout: docs
...
+
+A team owner can choose to prompt their teammates to upgrade to new Tilt release sooner than a week. If they visit their team's page on Tilt Cloud, they will find a button to "promote" the release to their team. This button tells Tilt Cloud to tell their team to use the latest version of Tilt instead of whatever it was previously telling them.
```
victorwuky, line 40:
```suggestion
A team owner can choose to prompt their teammates to upgrade to new Tilt release sooner than a week. If they visit their team's page on Tilt Cloud, they will find a button to "promote" the new version to their team. If they promote, all users (not just `Owner`s) will now get the version nudge corresponding to that latest Tilt version.
```

victorwuky:
@landism : How about we call this feature "Canary releases", or whatever @dbentley thinks makes sense. And write it down in the docs title here. That way we can easily refer to it when we share with customers and reinforce the feature name/functionality when when we have customer conversations.

docs/teams.md
landism: I included the phrase "default behavior" to get the broad idea across while indicating that there are exceptions. IIUC, your suggestion is trying to capture one of those exceptions (owners) but not another (manual promotions), while dropping the "default behavior" qualifier, which makes this statement not entirely true by itself.

Was your main goal to get the "non-Owner" exception in there? How would you feel about "the default behavior is that non-Owner users..."?

landism:
> @landism : How about we call this feature "Canary releases", or whatever @dbentley thinks makes sense. And write it down in the docs title here. That way we can easily refer to it when we share with customers and reinforce the feature name/functionality when when we have customer conversations.

changed it in the title and one other place

docs/teams.md
victorwuky: The latest version looks good.

PR #563, opened 2020-05-04
docs: add hover anchor links to doc <h2>s

landism:
Lots of our docs have a lot of `<h2>`s that I've wanted to link directly to, but could not do so without inspecting html and constructing a link. Now they get hover anchor links like we already had on `<h3>`s!

e.g.

image

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -62,5 +62,3 @@ Tiltfiles are written in _Starlark_, a dialect of Python. For more information o
 ---
 
 {% include api/classes.html %}
-
-<script src="/assets/js/links.js" async></script>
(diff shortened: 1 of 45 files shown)


docs/api.md
```diff
@@ -62,3 +62,5 @@ Tiltfiles are written in _Starlark_, a dialect of Python. For more information o
 ---
 
 {% include api/classes.html %}
+
+<script src="/assets/js/links.js" async></script>
```
nicks, line 66: have you considered adding this to 
https://github.com/windmilleng/tilt.build/blob/master/src/_layouts/docs.html
?

landism: Nope, I didn't know that existed and hadn't spent the two seconds thinking about it to realize it must exist. That seems much better, thanks!

PR #570, opened 2020-05-07
Specify version settings check updates false overrides canary nudges

victorwuky:


docs/teams.md
```diff
@@ -35,7 +35,7 @@ Tilt's web UI will nudge users to upgrade when a new release is available, at th
 
 When running Tilt with a Tiltfile that specifies a team id, the default behavior is that users are only notified about Tilt releases once they're one week old<sup>[\*](#minimum-suggested-version)</sup>.
 
-However, the team's *owners* are notified of Tilt releases as soon as they come out, giving them an opportunity to canary the new version before the rest of their team. If no action is taken, Tilt will begin notifying non-`Owner`s of a release once it's a week old.
+However, the team's *owners* are notified of Tilt releases as soon as they come out, giving them an opportunity to canary the new version before the rest of their team. If no action is taken, Tilt will begin notifying non-`Owner`s of a release once it's a week old. (If [`version_settings(check_updates=False)`](./api.html#api.version_settings) is set in the Tilfile, users don't get notified, regardless of their role.)
```
landism, line 38:
```suggestion
However, the team's *owners* are notified of Tilt releases as soon as they come out, giving them an opportunity to canary the new version before the rest of their team. If no action is taken, Tilt will begin notifying non-`Owner`s of a release once it's a week old. (If [`version_settings(check_updates=False)`](./api.html#api.version_settings) is set in the Tiltfile, users don't get notified, regardless of their role.)
```

PR #572, opened 2020-05-08
Help teams not yet on Kubernetes, adopt Tilt

victorwuky:


victorwuky:
@nicks : Some problems that @dbentley and I have talked about that are slowing down customer adoption of Tilt:

- Customers are developing big, long platform plans to revamp their development and production deploy infrastructures. Tilt (and Tilt adoption) is one component of many pieces here. These plans are inherently risky and take a long time, and Tilt is taking on that risk by being part of those plans.
- Teams have big projects with many services. It may take a lot of work to adopt Tilt all at once.

We talked about creating some artifacts to help our own account managers guide customers. But these can also be consumed by customers themselves. The purpose of this document I added is to encourage:

- Development teams to adopt Tilt as soon as possible, even hopefully ahead of any production Kubernetes plans have been officially established / rolled out.
- Development teams to adopt Tilt incrementally (one service at a time).

The document here is just some bullet points and serves as an outline. What do you folks think? Worth pursuing?

I am a bit worried that asking teams to adopt Tilt incrementally doesn't offer much value. Or maybe it does?

docs/kubernetes_migration_guide.md
```diff
@@ -0,0 +1,43 @@
...
+_Document outline_
+
+## Migrating to Kubernetes
+- Are you migrating to Kubernetes?
+- This guide is for you.
+- Tilt helps developments teams do development, that are already deploying to Kubernetes now, or are planning to do so in the future.
+- Many teams focus on the production aspect of Kubernetes. They often don't realize that it has downstream impacts on development. This guides puts both together in perpsective.
+
+## Production migration
+- Here are some resources on production migration of Kubernetes.
```
dbentley, line 15: I'd even say "There are lots of people who will do this, so we're not going to." instead of trying to maintain something that's not our competency.

docs/kubernetes_migration_guide.md
```diff
@@ -0,0 +1,43 @@
...
+- Tilt is not designed for deploying to production.
+
+## Development migration
+- We've seen many teams migrate to Kubernetes for production, without much thought of the impacts on development.
+- In the worst case, development teams have needed to scramble to adjust existing development workflows to accommodate Kubernetes in production, on strict timelines.
+- We recommend that you have a development migration plan in conjunction with your production migration plan.
+- It's a best practice to decouple (as much as possible) the dev plan and the prod plan. Lots of moving pieces and inherent risk to get things working altogether at once.
+- If you are the dev team, there's no need to wait for production migration. You can do development migration first, in short:
```
dbentley, line 23: This says "use Tilt for dev on Kubernetes before using prod on Kubernetes".

I think the more important message is "use Tilt for dev on non-Kubernetes before doing dev on Kubernetes. This makes it easier to switch dev to Kubernetes in the future."

(This is only relevant to people who have non-Kubernetes dev already).

victorwuky: Ah this is interesting @dbentley . Tilt can be used without k8s. (But needs at least Docker right? I.e. replacing docker-compose? https://docs.tilt.dev/docker_compose.html). So the migration path would be:

- Team is using Docker containers and docker-compose
- Team uses Docker containers and Tilt
- Team uses Docker containers and local Kubernetes cluster and Tilt
- Production deploy uses Kubernetes

dbentley: You can use Tilt with bare processes with local_resource. https://blog.tilt.dev/2020/02/12/local-dev.html

victorwuky: Oh great! There's so much that can be done in terms of guides / example projects here. Will make sure to get Nick's take before more revisions here.

victorwuky:
@dbentley : Spent more time thinking about this over the weekend and how to better position it. I like the framing of "Tilt Adoption Guide" much better. 

- We want to help help companies who have discovered Tilt, roll it out to their users more quickly.
- Often there is one initial champion of Tilt within an org, and they speak to us and we have a relationship with this one person.
- We want to help that person roll out Tilt, whether that person is more on the operations side, or dev ex side.
- We've chatted about helping dev teams and providing more incentive for dev teams and dev personas to adopt Tilt.
- This doc is precisely targeted at helping a dev team lead or a dev team advocate roll out Tilt incrementally. It is focused on emphasizing immediate developer value (for Frieda!) and how to achieve this value with low effort.

nicks:
At a high-level, I'm concerned that this doc recommends an approach that prioritizes what we want ("get everyone to install Tilt") over what our users find valuable. It doesn't fulfill our promise to recommend best practices that we've seen work well across different teams.

When I look at the anonymized metrics, very few repos use local_resource if they're not already using docker_compose or k8s_yaml (it's something like 6%)

If I was a dev ex owner, and a guide told me to get on my team to install a tool before I had evaluated it, I simply would not trust that tool.

That said, I'm fine with saying that this is a provisional/suggested message that we want to test out, I'm just skeptical

victorwuky:
@nicks : The doc is much more focused now. Feel free to parse again.

PR #573, opened 2020-05-08
Enterprise page fixes

hyu:
Add the Enterprise icon
And fix the Calendly link on the Enterprise page

Screen Shot 2020-05-08 at 6 47 17 PM


src/_data/header.yml
```diff
@@ -21,8 +21,8 @@
   icon: blog
   mobileIcon: blog
   blogurl: true
-# - name: Enterprise
-#   icon: enterprise
-#   mobileIcon: enterprise
-#   href: enterprise
-#   landingurl: true
+- name: Enterprise
```
nicks, line 24: giphy

hyu: You can tell we're serious because there's a little briefcase icon!

PR #574, opened 2020-05-11
api: document secret_settings

maiamcc:
docs for [tilt#3313](https://github.com/windmilleng/tilt/pull/3313)

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -805,8 +805,7 @@ def secret_settings(disable_scrub: bool = False) -> None:
   'mysecurepassword', Tilt redacts this string if ever it appears in the logs,
   to prevent users from accidentally sharing sensitive information in snapshots etc.
 
-  Currently ``secret_settings`` only turns secret scrubbing on and off. Expect more
-  settings to be configurable from this function soon.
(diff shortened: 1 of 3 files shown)


api/api.py
```diff
@@ -798,6 +798,21 @@ def struct(**kwargs) -> Any:
...
   """
 
+
+def secret_settings(disable_scrub: bool = False) -> None:
+  """Configures Tilt's handling of Kubernetes Secrets. By default, Tilt scrubs
+  the text of any Secrets from the logs; e.g. if Tilt applies a Secret with contents
+  'mysecurepassword', Tilt redacts this string if ever it appears in the logs,
+  to prevent users from accidentally sharing sensitive information in snapshots etc.
+
+  Currently ``secret_settings`` only turns secret scrubbing on and off. Expect more
```
landism, line 808: I think it's good to avoid relative time statements ("currently", "soon") in undated documents, since anyone reading this has no idea how long ago this was written. It's also making a commitment I'm not sure we want to make? Maybe instead of "expect more", "if you have more sophisticated needs, let us know"?

nicks:
the release is out, merge away!

PR #578, opened 2020-05-13
api: add docs for batch commands

nicks:
Hello @jazzdan, @landism,

Please review the following commits I made in branch nicks/bat:

e781a7f71b5723c6a6be8e7e796b41942927a0a2 (2020-05-13 18:28:56 -0400)
api: add docs for batch commands

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -344,12 +344,14 @@ def load(path: str, *args):
     hi() # prints "Hello world!"
   """
 
-def local(command: str, quiet: bool = False) -> Blob:
+def local(command: str, quiet: bool = False, command_bat: str = "") -> Blob:
   """Runs a command on the *host* machine, waits for it to finish, and returns its stdout as a ``Blob``
 
   Args:
-    command: Command to run.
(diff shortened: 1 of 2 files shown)


api/api.py
```diff
@@ -344,12 +344,14 @@ def load(path: str, *args):
...
 
-def local(command: str, quiet: bool = False) -> Blob:
+def local(command: str, quiet: bool = False, command_bat: str = "") -> Blob:
   """Runs a command on the *host* machine, waits for it to finish, and returns its stdout as a ``Blob``
 
   Args:
     command: Command to run.
     quiet: If set to True, skips printing output to log.
+    command_bat: The command to run, expressed as a Windows batch command executed
+      with ``cmd /S /C``. Takes precedence over the ``command`` parameter on Windows.
```
jazzdan, line 354: What happens if you specify it in a non-Windows context?

jazzdan: What happens if you pass `command` in Windows?

nicks: clarified this a bit

PR #585, opened 2020-05-15
guides: update guides to reflect restart_process extension instead of entr hack for restarting

maiamcc:
Hello @nicks, @jazzdan,

Please review the following commits I made in branch maiamcc/update-ex-guides-restart_process:

061cf97b5ea118d27de643a7f832360c3815d77e (2020-05-15 19:22:23 -0400)
guides: update guides to reflect restart_process extension instead of entr hack for restarting

### DIFF
--- a/docs/example_csharp.md
+++ b/docs/example_csharp.md
@@ -223,9 +223,9 @@ The first thing to notice is the `live_update` parameter, which consists of one 
 This copies the build output from the `out` directory into the container.
 
 After syncing the files, we want to re-execute our updated binary. We accomplish this with the
-[`restart_process` extension](https://github.com/windmilleng/tilt-extensions/tree/master/restart_process)
-(for more on extensions, [see the docs](/extensions.html)).
-In particular, we swap out our `docker_build` call for a `docker_build_with_restart` call; it's
+[`restart_process` extension](https://github.com/windmilleng/tilt-extensions/tree/master/restart_process),
+which we imported with the `load` call on the first line. (For more on extensions, [see the docs](/extensions.html).)
+We swap out our `docker_build` call for function we imported, `docker_build_with_restart`: it's
 almost exactly the same as `docker_build`, only it knows to restart our process at the end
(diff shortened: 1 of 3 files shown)


nicks:
would love to have @ellenkorbes review this instead of me. 

How we introduce extensions here is going to speak a lot to how we introduce extensions to developers in general

Like, an alternative way to do this would be to break it up into two separate steps: "here's how you'd wire live_update with just run() calls. But the recommended way to do it is with an extension. extensions are a great way to share best practices with other people. here's how you do it"

But I could be convinced that this is too much for an initial best practices guide

docs/example_csharp.md
```diff
@@ -206,23 +206,28 @@ With Tilt, we can skip all of these steps, and instead [live_update](live_update
...
-We've added a `live_update` parameter to `docker_build()` with a `sync` step. This copies the build output from the `out` directory into container.
+The first thing to notice is the `live_update` parameter, which consists of one `sync` step.
+This copies the build output from the `out` directory into the container.
 
-We've also added a new parameter: `entrypoint='find . | entr -r dotnet hello-tilt.dll'`.
-
-`entr` is a tool that automatically restarts a shell command whenever the watched
-file changes. This command restarts our server every time the pod is updated.
+After syncing the files, we want to re-execute our updated binary. We accomplish this with the
+[`restart_process` extension](https://github.com/windmilleng/tilt-extensions/tree/master/restart_process)
```
ellenkorbes, line 226: SGTM, I'd just add a short remark about what the `load(ext)` call is. E.g.

```suggestion
[`restart_process` extension](https://github.com/windmilleng/tilt-extensions/tree/master/restart_process), which we imported on the first line
```

maiamcc: ooh good call, ty

maiamcc:
@ellenkorbes (or whoever) can I get a ✅ ?

ellenkorbes:
There!

PR #586, opened 2020-05-18
docs: entrypoint can be passed as string or argv

maiamcc:
this wasn't documented whoooops. Good catch @nicks

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -78,7 +78,7 @@ def restart_container() -> LiveUpdateStep:
   """
   pass
 
-def docker_build(ref: str, context: str, build_args: Dict[str, str] = {}, dockerfile: str = "Dockerfile", dockerfile_contents: Union[str, Blob] = "", live_update: List[LiveUpdateStep]=[], match_in_env_vars: bool = False, ignore: Union[str, List[str]] = [], only: Union[str, List[str]] = [], entrypoint: str = "", target: str = "", ssh: Union[str, List[str]] = "", network: str = "", secret: Union[str, List[str]] = "", extra_tag: Union[str, List[str]] = "", container_args: List[str] = None) -> None:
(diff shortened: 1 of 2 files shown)


maiamcc:
Yep, commands passed to [local](https://github.com/tilt-dev/tilt/blob/b81141ff0587af27e2b9978516ba22046f60fe9a/internal/tiltfile/files.go#L35) or to [local_resource](https://github.com/tilt-dev/tilt/blob/b81141ff0587af27e2b9978516ba22046f60fe9a/internal/tiltfile/local_resource.go#L72) both get run thru `ValueToHostCmd`. Docs updated!

api/api.py
```diff
@@ -344,11 +344,11 @@ def load(path: str, *args):
     hi() # prints "Hello world!"
   """
 
-def local(command: str, quiet: bool = False, command_bat: str = "") -> Blob:
+def local(command: Union[str, List[str]], quiet: bool = False, command_bat: str = "") -> Blob:
```
maiamcc, line 347: @nicks afaict the point of `command_bat` is to make sure the command gets executed as `["cmd", "/S", "/C", cmd]`, right? Seems to me that if you've got a windows Tilt install, if you pass a string to `command` it goes through the same codepath as `command_bat` -- are these two actually redundant or am i missing something?

nicks: here's an example that might help:
https://github.com/tilt-dev/tilt/blob/7dadfff0fa68c929932a20e7ecf72444b0595378/internal/tiltfile/tiltfile_test.go#L287

This kind of API isn't totally unexpected if you're used to other cross-platform tools that need to run scripts on windows:
https://docs.bazel.build/versions/master/be/general.html#genrule.cmd_bat

maiamcc: aaaah I see, it's not because you want the same command you passed to `cmd`, but handled differently; it's because you want to, within a single tiltfile, provide one command to run if it's osx/linux and one command if it's windows. 👍 

Still, it's the case that `command_bat` values go through the same `valueToCmdHelper` as our other commands, and so can be passed as either string or array. do you want to / should I update the docs to reflect that?

PR #594, opened 2020-05-26
New Header for A/B Testing

hyu:
Per a suggestion from @nicks, I created a new header which can be A/B tested against the existing header. (Rather than replacing the existing header in this PR.) 

To test, make changes to default.html like so:

```
  <body class="body body--{{ page.layout }}">

    {%- include header2.html -%}

    <main class="page-content--header2" aria-label="Content">
      {{ content }}
    </main>
...
```

## Desktop
2020-05-25 - header

## Mobile
2020-05-25 - header mobile
2020-05-25 - header mobile menu


hyu:
Hmm, I see failing deploys here...

~~~
Conversion error: Jekyll::Converters::Scss encountered an error while converting 'assets/css/style.scss':
Invalid CSS after ";": expected "}", was "" on line 149
~~~

style.scss is compiled from all the separate SCSS files. When I run locally, Tilt doesn't show any errors. And when I inspect the Tilt.dev in my dev cluster, line 149 points me to a part of the CSS that looks fine. 

Any idea how I can pinpoint where this error is actually happening?

src/_sass/tilt.scss
```diff
@@ -116,8 +119,22 @@ $on-docs-mobile:    $header-width + $spacing-unit;
 
 @mixin relative-font-size($ratio) {
   font-size: $body-font-size * $ratio;
+@mixin mobileAndBelow {
```
nicks, line 122: i don't think this will parse? you need to close the relative-font-size mixin before you start a new one

src/_includes/header2.html
```diff
@@ -0,0 +1,53 @@
...
+
+<header class="siteHeader2">
+  <div class="siteHeader2-logoAndMenuButton">
+    {% if page.layout == 'landing' %}
+    <!-- For the "Guide To App Devs" - /welcome_to_tilt.html -->
+    <a class="imageLink imageLink--logo-imagemark" href="{{site.landingurl}}/">
+      {% svg assets/svg/logo-imagemark.svg %}
+    </a>
+    {% elsif page.layout == 'home' %}
+    {% svg assets/svg/Tilt-logo.svg class="siteHeader2-logoLink-svg" %}
```
nicks, line 11: for my understanding, is there a reason you created a new svg logo? we already have 3 logo-wordmark.svg's - one with an alpha channel, one with solid colors, and one for animations

src/_includes/header2.html
```diff
@@ -0,0 +1,53 @@
...
+          {% assign href = (href | prepend: "/" | prepend: site.landingurl) %}
+        {% elsif link.docsurl %}
+          {% assign href = (href | prepend: "/" | prepend: site.docsurl) %}
+        {% elsif link.blogurl %}
+          {% assign href = (href | prepend: "/" | prepend: site.blogurl) %}
+        {% endif %}
+
+        {% if link.github %}
+          <li class="siteHeader2-navItem">
+            <a class="siteHeader2-navLink--github" href="{{href}}">
```
nicks, line 37: usually, in BEM, you only have the modifier class on the element if you also have the base class. So this would be class="siteHeader2-navLink siteHeader2-navLink--github"
http://getbem.com/naming/

src/_layouts/default.html
```diff
@@ -5,9 +5,9 @@
 
   <body class="body body--{{ page.layout }}">
 
-    {%- include header.html -%}
+    {%- include header2.html -%}
```
nicks, line 8: The include directive is run when Jekyll generate the HTML. Not when we're serving the website. So there's no way to A/B test this as-is.

One way to do this is to render BOTH headers and display:none one when a css class is set on the body. This is pretty easy to toggle with optimize

src/_data/header2.yml
```diff
@@ -0,0 +1,16 @@
+- name: Enterprise
+  href: enterprise
+  primary: true
+  landingurl: true
+- name: Docs
+  href:
+  docsurl: true
+- name: GitHub
+  href: https://github.com/tilt-dev/tilt
+  githubStars: 3.5k
```
nicks, line 10: hahah nice work

src/_sass/header2.scss
```diff
@@ -0,0 +1,204 @@
...
+
+    li + li {
+      margin-left: $spacing-unit * 0.5;
+    }
+  }
+}
+
+@mixin button--siteHeader {
+  font-family: $font-sansSerif;
+  padding-top: $spacing-unit * 0.1;
```
nicks, line 64: i don't think spacing-unit is divisible by 10? we usually do these in powers of 2 (unit / 8, unit / 4, etc)

src/_sass/tilt.scss
```diff
@@ -73,6 +73,9 @@ $color-blue-lightest: #98E8ED;
 $color-green-light: #70D37B;
 $opacity-light: 0.64;
 
+// Updated 2020-05
+$color-gray: #586E75;
```
nicks, line 77: i think it's asking for trouble to have both $color-gray and $color-grey, and make them different colors

src/_sass/header2.scss
```diff
@@ -0,0 +1,204 @@
+$header2-width:    1080px;
+
+$color-white: #ffffff;
```
nicks, line 3: can you move this into the existing color scale?

nicks:
as an aside, we should add a mechanism to the tiltfile to test for the parse error locally! yay dogfooding!!

src/_data/header2.yml
hyu: Haha, I figure I could make a new story to make this properly update...

src/_layouts/default.html
hyu: Ohhhh, that makes sense now. Will do...

src/_includes/header2.html
hyu: Haha, this was kind of lazy of me. The new layout has a larger logo. Since default height and width are part of the SVG, I preferred to export a new SVG instead of using CSS to alter the existing dimensions. (Which would have meant hardcoding a seemingly arbitrary width in CSS.)

If we go with this new header, we also won't need the solid vs anim versions, since I removed the animation effect. (It was cute but I thought a bit distracting.) I also think people sometimes download logos for use from websites, so I named it "Tilt-logo" so it's a bit more of a branded product.

Overall, my assumption is that we'll make a decision whether to go with the new layout or revert to the existing layout, so I wanted to make it easier to delete one set or another rather than re-use existing elements.

src/_sass/tilt.scss
hyu: OOPS — this was an obvious error that was caused by a hasty merge. Then I missed it because I was looking in the wrong branch. THAT'S why I couldn't see the error in the CSS even though it was super obvious and actually readily caught by Tilt.

src/_layouts/default.html
hyu: Or, actually, I could use your help to make that small change once I merge this PR, just so I can continue focusing on the Landing Page layout.

src/_sass/tilt.scss
hyu: Oops, I meant to have this inside header2.scss...

src/_layouts/default.html
nicks: ok

src/_sass/header2.scss
hyu: I had it in the shared tilt.scss file, but I moved them into this header2.scss because there's actually a subtly different color scale which I'll be implementing along with the refresh of the entire Landing Page. I wanted to make it clear that this is not a part of the original layout (aka, what's in tilt.scss). 

Is it OK if I leave these inside header2 for now, and then further pull these out in a shared CSS which is clearly part of the new layout in the forthcoming Landing Page PR?

src/_includes/header2.html
hyu: Oh! Duly noted. I think I was used to other BEM-adjacent systems that were a bit more loose with the use of modifiers.

src/_sass/header2.scss
hyu: Half pixels do render, but I agree that it's best to keep them integers. Thanks for the note!

src/_sass/header2.scss
nicks: ok. if we're going to do a new color scale, can we namespace them in some way? e.g., $color2-...

I'm pretty sure that moving them into header2.scss doesn't solve the problem? There's no file-level scope in SASS. So it doesn't really matter what file they're in. Setting a global variable just overwrites the other one for future declarations, which makes this harder to reason about

hyu: Good point! I know that there's no file-level scope in SASS, but I hoped that keeping them in header2.scss would show they weren't for general use. That said, it wasn't explicit enough, so I added a prefix ("header2") and also added a comment.

hyu:
Ok, made a bunch of suggested changes. @nicks — can you take another look and see if you can approve?

hyu:
@nicks — Yes! I reverted the changes in default.html.

PR #598, opened 2020-05-27
src: make features list interactive

jazzdan:


src/index2.md
```diff
@@ -52,55 +52,55 @@ layout: home2
...
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar1">Magic UI</button></li>
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar2">live_update</button></li>
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar2">Code in Flow</button></li>
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar2">Fastr Workflow</button></li>
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3">Snapshots</button></li>
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3">Built-In Best Practices</button></li>
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3">Painless Onboarding</button></li>
-    <li><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3">Quantified Dev Experience</button></li>
+    <li class="Home2-features-navItem"><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar1" href="#1">Very Holistic</button></li>
```
hyu, line 55: Since you're using a `href` attribute, you can use `a` instead of `button`

src/index2.md
```diff
@@ -52,55 +52,55 @@ layout: home2
...
+    <li class="Home2-features-navItem"><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar2" href="#5">Code in Flow</button></li>
+    <li class="Home2-features-navItem"><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar2" href="#6">Faster Workflow</button></li>
+    <li class="Home2-features-navItem"><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3" href="#7">Snapshots</button></li>
+    <li class="Home2-features-navItem"><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3" href="#8">Built-In Best Practices</button></li>
+    <li class="Home2-features-navItem"><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3" href="#9">Painless Onboarding</button></li>
+    <li class="Home2-features-navItem"><button class="Home2-features-navItem-button Home2-features-navItem-button--pillar3" href="#10">Quantified Dev Experience</button></li>
   </ul>
   <ul class="Home2-features-contentList">
-    <li>
+    <li class="Home2-features-contentItem" id="1">
```
hyu, line 67: Can the id name be more descriptive? Like, hm..."Home2-feature-item[1]"?

src/_sass/home2.scss
```diff
@@ -244,6 +244,8 @@
 // Features --------------------------------------------------
 .Home2-features {
   margin-bottom: $spacing-unit;
+  // TODO(dmiller): there's probably a better way to do this
+  height: 300px;
```
hyu, line 248: To make it roughly closer to the height of the menu...

```suggestion
  height: 400px;
```

src/_sass/home2.scss
```diff
@@ -295,6 +297,7 @@
 .Home2-features-contentList {
   list-style: none;
   margin-left: 0;
+  overflow: scroll;
```
hyu, line 300: I didn't see a horizontal scroll when I played around in browser, but just to ensure there's only vertical scroll...

```suggestion
  overflow-y: scroll;
```

jazzdan:
@hyu changes made!

nicks:
@hyu PTAL?

I ended up throwing out most of the old approach. I feel like it copied a lot of code from the codepen that didn't make sense for our use-case -- everything we're trying to do here is supported natively by modern browsers

I also moved the pillars to a data file, since it seems likely we'll change the layout a few more times (especially as we go to mobile) and will need to duplicate some of the content multiple times in the html

hyu:
Ooh, I like this approach @nicks. Good idea to put the pillars in a data file so it's easier to re-use.

I just tested this in both Firefox and Chrome, though, and it feels like it doesn't scroll to the position I would expect. If I click on "Snapshots", for example, I'd want to see it aligned to the top. It is within the viewport, though. What are you seeing?

nicks:
Here's what I see after I click "Snapshot". Is that what you expect?

Screenshot from 2020-05-28 13-15-38

nicks:
I can imagine at least 4 interactions that might make sense to me:
1) Scroll the inner div so that the content item lines up with the top of the features box (the behavior in this PR)
2) Scroll the inner div and the global window so that the content item lines up to the top of the viewport (like in https://codepen.io/eksch/pen/xwdOeK/)
3) Scroll the inner div so that the content item lines up with the nav item (like what's in the figma mock)
4) Scroll the global window and positiion: sticky the navigational box (like in https://css-tricks.com/position-sticky-2/)

nicks:
(part of my confusion is that the ticket points to https://codepen.io/eksch/pen/xwdOeK/ which is closer to (4), tho the ticket also says there should be an inner scrolling div)

PR #599, opened 2020-05-28
home: Testimonials and Resources design

hyu:
Cleaned up Testimonials and Resources for mobile and desktop. No interaction yet, though!

Screen Shot 2020-05-27 at 9 13 52 PM

Screen Shot 2020-05-27 at 9 16 14 PM

--------------------------------------------------

Screen Shot 2020-05-28 at 9 34 11 AM
Screen Shot 2020-05-28 at 9 34 24 AM


src/_sass/home2.scss
```diff
@@ -352,8 +454,70 @@
...
+  line-height: $tilt2-line-height-body;
+  padding-top: $spacing-unit / 4;
+  padding-bottom: $spacing-unit / 4;
+  border-radius: $tilt2-border-radius;
+  border: 1px solid $color-blue;
+  text-align: center;
+  color: $color-blue;
+
+  &:hover {
+    color: $color-white;
```
nicks, line 494: should this have a `transition` on color/background-color?

src/_sass/tilt2.scss
```diff
@@ -3,35 +3,44 @@
...
 // TYPOGRAPHY
 // --------------------------------------------------
-$tilt2-font-sans: 'Noto Sans', noto-sans, sans-serif;
+$tilt2-font-sans: noto-sans, sans-serif;
+$tilt2-font-serif: merriweather, serif;
 $tilt2-font-mono: 'Fira Code', monospace;
 
+$tilt2-font-weight-sans-light: 300;
+$tilt2-font-weight-sans-bold: 700;
+$tilt2-font-weight-sans-black: 900;
```
nicks, line 25: I'm generally skeptical of adding a new font weight that we only use in one place. It seems messy from both a design perspective (because it, by definition, breaks out of the existing design system) and an eng perspective (because we have to download a new font for each new weight).

src/assets/svg/resources-contact.svg
```diff
@@ -0,0 +1 @@
+<svg fill="none" height="41" viewBox="0 0 41 41" width="41" xmlns="http://www.w3.org/2000/svg"><path d="m38.75 8.75h-4v18h-26v4c0 1.1.9 2 2 2h22l8 8v-30c0-1.1-.9-2-2-2zm-8 12v-18c0-1.1-.9-2-2-2h-26c-1.1 0-2 .9-2 2v28l8-8h20c1.1 0 2-.9 2-2z" fill="#03c7d3"/></svg>
```
nicks, line 1: for what it's worth, I've been trying to give all our SVGs CSS classes (like fillStd) instead of hard-coded colors. It makes it much easier to do hover effects and align them with our color scale.

src/_sass/tilt2.scss
hyu: I'm not 100% sure I'm interpreting this comment correctly, but I think you're saying that we should have fewer font weights. I tried to make the design work with just two sans-serif font weights — light and bold. But I think the black really makes the hero heading a lot more impactful. It's the first thing you see, and the effect is quite different (and worse, I think) in bold.

I'm re-reading this comment, though, and I could interpret it to mean that you want the font-weights to be more closely associated with each type of font? i.e., I only want 300, 700, and 900 to be available for use with our sans font "Noto Sans" (and not our mono or serif font). If that's the case, I would actually like to make that usage more clear in the shared SCSS but I haven't yet gotten around to it...

hyu: Oh, I've also seen a neat trick where you can download only the few characters you need from a Google Font. Like so:

~~~
http://fonts.googleapis.com/css?family=Inconsolata&text=MyHeader
~~~

But the Google Fonts version of Noto Sans lacks the "black" weight we want. So I'm using the Adobe Typekit version of Noto Sans. And the `&text` trick doesn't seem to work there.

nicks: more the first one. I think the way we've addressed this in the past is to use a darker color rather than a bolder font

i don't feel strongly about it though, it's your decision

nicks: (tho your second reading is also a good point that I wish I had said! We should probably have font mixins that group family + weight together; you almost never want to set them independently)

src/_sass/home2.scss
hyu: Yes! Good note.

src/assets/svg/resources-contact.svg
hyu: Thanks for catching this! I agree with you and this was unintentional.

PR #601, opened 2020-05-28
update docs to reference restart_process extension instead of restart_container live update step when possible [ch5945]

maiamcc:
Hello @landism, @nicks,

Please review the following commits I made in branch maiamcc/restart-process-reference:

43451058dfc04fdcffca4ef681902f611c20d380 (2020-05-28 14:47:00 -0400)
update docs to reference restart_process extension instead of restart_container live update step when possible [ch5945]


maiamcc:
This PR addresses API docs and Live Update Reference. Future PRs will tackle the Live Update Tutorial and a fuller docs audit

docs/live_update_reference.md
```diff
@@ -113,13 +111,39 @@ docker_build('my-img', '.', live_update=[
 2. change to `./src/web/yarn.lock` => run the Live Update; run `setup.sh`; run `cd /app/web && yarn install`, because this file matches that command's trigger
 3. change to `./configs/foo.yaml` => whoops, this file doesn't match any `sync` steps! Even though it matches a trigger (for the third `run`), we won't do a Live Update for this change; instead, we do a full Docker build (see notes on `sync`, above, for what changes trigger a Live Update vs. a full build + deploy).
 
-### [`restart_container()`](api.html#api.restart_container)
+## Rerunning your Process
```
nicks, line 114: I don't know what "rerunning" means here. let's use "restart" to be consistent with how we use it elsewhere.

docs/live_update_reference.md
```diff
@@ -113,13 +111,39 @@ docker_build('my-img', '.', live_update=[
...
 
-This step is optional. If you have a `restart_container` step, it must come at the very end of your list of Live Update steps. When this step is present, it tells Tilt to restart the container after all the files have been sync'd, runs have been executed, etc. In practice, this means that the container re-executes its `ENTRYPOINT` within the changed filesystem.
+Some apps or invocations thereof (e.g. Javascript apps run via `nodemon`, or Flask apps run in debug mode) detect and incorporate code changes without needing to restart. For other apps, though, you'll need to re-execute them for changes to take effect.
```
nicks, line 116: At a high-level, this uses a lot of explanation and justification words like
- "workaround"
- "exception"
- "most setups"

This is going to be difficult for most users to understand. It means that they have to know what the whole decision tree of setups looks like to make a decision. 

There are a couple common patterns to help users jump directly to what they need, e.g.,

Clearly lay out the decision tree for them. like:

```
## Node

Use entrypoint=nodemon

## Python

Use entrypoint=flask dev

## Custom_build

Do X Y Z
```

an alternate way is to break it up into easy/intermediate/advanced, like the cluster guide does

@ellenkorbes might have good suggestions here too.

nicks: (and to be clear, feel free to merge with what you like, and we can go back and iterate later, this is at least a big improvement on what we have now)

maiamcc: yep, that's really good feedback. going to merge what I have for the sake of moving this forward, but we should def come back and iterate here.

PR #609, opened 2020-05-29
Add video to python example

victorwuky:


victorwuky:
@ellenkorbes : I think we should add your video to the doc, especially if it's a great companion video, which it is! (Don't know where it should belong though. Feel free to move around if you agree we should add it.)

ellenkorbes:
Wait, why did you close the PR before merging?

victorwuky:
@ellenkorbes: I didn’t hear from you so I figured this wasn’t important to add should we add this?

ellenkorbes:
Yes. Probably here? https://docs.tilt.dev/example_python.html

ellenkorbes:
Yes! I'll review this tomorrow.

PR #618, opened 2020-06-02
docs: deprecate k8s_resource_assembly_version

jazzdan:


src/_includes/api/functions.html
```diff
@@ -5525,7 +5525,7 @@
                    str
                   </span>
                  </code>
-                 ) &#x2013; Path to the file locally (absolute, or relative to the location of the Tiltfile).
```
nicks, line 5528: this happened on my machine too and I honestly couldn't figure out why it was happening

PR #642, opened 2020-06-18
site: CTA links to install page

victorwuky:


victorwuky:
@hyu : I think we should link to the install page. Please let me know if that wasn’t the original intention and we can discuss.

hyu:
@victorwuky — I agree with you! I think the Install page is better because we can track it. 

(That said, I am also itching to tackle the design of our Docs at some point so it better conveys the feeling that getting started with Tilt will be straightforward and even pleasant. Currently there are issues with color, typography, and whitespace there.)

PR #644, opened 2020-06-18
site: CTA button classes for Google Analytics tracking

victorwuky:
Add classes to the CTA buttons so that can create targeted triggers in Google Tag Manager, in order to track clicks for Google Analytics.


nicks:
can you use a class name prefix to indicate what these are for?

victorwuky:
@nicks : Does the latest work?

PR #655, opened 2020-06-26
docs: Warning about mistakenly creating public snapshots

victorwuky:


docs/snapshots.md
```diff
@@ -18,6 +18,8 @@ Visit [https://cloud.tilt.dev/snapshot/Afygp-8LJ4vRmVdGtHU=](https://cloud.tilt.
 
 Inside Tilt, click the `Create a snapshot` button at the top. If you haven't yet connected Tilt to Tilt Cloud, you'll be prompted to do so (using your GitHub account to sign in). Click `GET LINK` to create the snapshot. Click `OPEN` and Tilt will open the snapshot in Tilt Cloud in a new browser tab. Copy the URL of the snapshot and share it with other team members (via Slack or email, for example). The snapshot is only accessible to Tilt Cloud signed-in users who are a member (or owner) of the team
 
+**Important**: If you get an error when you first run `tilt up`, Tilt won't understand `set_team`. If you then take a snapshot, it will be a _public_ snapshot. (See below for public snapshots.) To work around this problem, make sure there are no errors, e.g. by commenting out the offending lines in your Tiltfile. Tilt will then successfully parse `set_team` and register the team id in it. You can then uncomment any offending lines or otherwise introduce errors. Create a snapshot as normal and it will be a private team snapshot. We're working on a better experience to fix this problem.
```
landism, line 21:
```suggestion
**Important**: If the Tiltfile errors on its first execution when you run `tilt up`, Tilt won't understand `set_team`. If you then take a snapshot, it will be a _public_ snapshot. (See below for public snapshots.) To work around this problem, make sure there are no Tiltfile errors, e.g. by commenting out the offending lines in your Tiltfile. Tilt will then successfully parse `set_team` and start using the team id in it. You can then uncomment any offending lines or otherwise introduce errors. Create a snapshot as normal and it will be a private team snapshot. We're working on a better experience to fix this problem.
```

landism: changed "an error" at the start to make clear that, e.g., compiler or k8s errors are fine

replaced "register with "start using". I don't love "start using" either, but I worry "register" could be interpreted a la "register an account"

landism: maybe also a note along the lines of "You can confirm the snapshot will be private by seeing that the create snapshot dialog tells you it will be a team-private snapshot"?

victorwuky:
> maybe also a note along the lines of "You can confirm the snapshot will be private by seeing that the create snapshot dialog tells you it will be a team-private snapshot"?

@landism : This won't be available yet until https://github.com/tilt-dev/tilt/pull/3518 is merged and we have a new release right? So I'll be sure to add that here when that happens. But I'll definitely mention it when I send this to our customer separately.

PR #656, opened 2020-06-26
fixed nav items in What We Have in Store section

smombartz:
removed blue highlight (on macos chrome)
made sharp bottom corners only for mobile

### DIFF
--- a/src/_sass/home.scss
+++ b/src/_sass/home.scss
@@ -328,10 +328,13 @@ $Home-feature-navList-width: 35%;
   transition: color $anim-duration-short linear,
               background-color $anim-duration ease,
               border-color $anim-duration ease;
+  outline: none;
 
   .Home-features-navItem.is-active & {
-    border-bottom-left-radius: 0;
-    border-bottom-right-radius: 0;
+    @include mobileAndBelow {
+      border-bottom-left-radius: 0;
+      border-bottom-right-radius: 0;
+    }
   }
 }
 


src/_sass/home.scss
```diff
@@ -328,10 +328,13 @@ $Home-feature-navList-width: 35%;
   transition: color $anim-duration-short linear,
               background-color $anim-duration ease,
               border-color $anim-duration ease;
+  outline: none;
```
hyu, line 331: @smombartz — We shouldn't remove outline: none with no alternative option, since that [harms accessibility](https://medium.com/better-programming/a11y-never-remove-the-outlines-ee4efc7a9968). 

Can you remove this from this PR, for now? In our Slack DM, I had proposed that we could use the same "hover" design for focus, but I realized that doesn't really work. The better solution is something like [unfocus.js](https://raw.githubusercontent.com/nuclei/unfocus/master/dist/unfocus.js), which "removes the outline if you use the mouse, and if you use the keyboard, you’ll enable back all the outlines."

You could also potentially design an outline that you think works better with our branding. It should be distinct from "hover," since it serves a different purpose. (You should be able to navigate the entire page using your keyboard, and know which element is currently active...and that's annoying if instead of a consistent outline you get a different focus style depending on which UI element is active.)

smombartz: makes sense! in that case i think whats there is good enough.

PR #664, opened 2020-07-02
docs: how to configure a custom resource

nicks:
Hello @jazzdan,

Please review the following commits I made in branch nicks/frameworks:

c70c6e08e7a5982d5e193e72d9a98268bee75e1f (2020-07-02 19:12:06 -0400)
docs: how to configure a custom resource

### DIFF
--- a/docs/custom_resource.md
+++ b/docs/custom_resource.md
@@ -1,0 +1,84 @@
+---
+title: "Custom Resource Definitions"
+description: "Configure your custom resources so that Tilt can monitor them."
+layout: docs
+---
+
+Kubernetes defines well-thought-out primitives like Deployments, Jobs, and
+StatefulSets for running your containers.
+
+Tilt knows how to inject images, monitor status, and fetch logs for these
+built-in resources.
+
+But your team may define your own custom resources! Or you may use a project
+like KubeDB that uses custom resources to operate a database.
+
+Tilt needs your help to recognize them.
+
+## Custom Resource Functions
+
+For any resource, Tilt needs to know 3 things:
+
+### Is this an indepdent resource?
+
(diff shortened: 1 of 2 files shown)


docs/custom_resource.md
```diff
@@ -0,0 +1,81 @@
...
+But your team may define your own custom resources! Or you may use a project
+like KubeDB that uses custom resources to operate a database.
+
+Tilt needs your help to recognize them.
+
+## Custom Resource Functions
+
+For any resource, Tilt needs to know 3 things:
+
+### Is this an indepdent resource?
```
jazzdan, line 22:
```suggestion
### Is this an independent resource?
```

But also, what happens if it's not an independent resource, what should you do then?

(I know the answer, I just think that the way that this is phrased leaves that question unanswered)

docs/custom_resource.md
```diff
@@ -0,0 +1,81 @@
...
+
+```
+k8s_resource(new_name='postgres',
+             extra_pod_selectors=[{'kubedb.com/name': 'quick-postgres'}])
+```
+
+See the [k8s_resource API docs](api.html#k8s_resource) for more detail, or look
+at [this
+example](https://github.com/tilt-dev/tilt-example-frameworks/blob/master/kubedb-postgres/Tiltfile)
+that selects the Postgres pods created by the KubeDB postgres operator.
```
jazzdan, line 66:
```suggestion
that selects the Postgres pods created by the KubeDB Postgres operator.
```

(We should just be consistent, doesn't matter to me how we capitalize it)

docs/custom_resource.md
nicks: done

docs/custom_resource.md
nicks: done

PR #673, opened 2020-07-14
docs: snapshot doc emphasizes snapshots first, privacy second

maiamcc:
Context: I’m pretty surprised at how much the existing snapshots doc feels like the feature it’s about is _private team snapshots_ rather than just… snapshots. (It seems like we wrote the doc to the engineering work we did rather than how the user experiences the feature.) It feels like we're putting the cart before the horse in terms of selling the privacy feature before we're necessarily sold the reader of this doc on snapshots themselves.

(The emphasis we're putting on private-ness also feels weird because... it seems like kind of a no-brainer that snapshots should be access controlled. it feels like a car ad bragging that your car has wheels?)

tangentially, I tried to make this doc less _explicitly_ Laura-directed -- the points in here are largely meant to resonate with Laura, but I took out some of the explicit "you, as the DevEx head" language so that this doc can also serve as a resource for a Frieda being told to use snapshots who's wondering what the heck it is. However, I feel less strongly about this point than I do about the above point.

not married to this approach but wanted to see what it would look like, would love thoughts. (I also have some ideas about how to make the screenshots/visual content more engaging with this new doc structure, but will hold off until I get people's thoughts on the structure/copy)


docs/snapshots.md
```diff
@@ -1,44 +1,47 @@
...
-
-**Important**: If the Tiltfile errors on its first execution when you run `tilt up`, and [`set_team`](https://docs.tilt.dev/api.html#api.set_team) has not been called yet, Tilt will not understand `set_team`. If you then take a snapshot, it will be a _public_ snapshot. (See below for public snapshots.) To work around this problem, make sure to put `set_team` as the first line in your Tiltfile. Tilt will then successfully parse `set_team` and start using the team id in it. Create a snapshot as normal and it will be a private team snapshot.
+## Creating and Sharing Snapshots
+Inside Tilt, click the `Create a snapshot` button at the top. If you haven't yet connected Tilt to Tilt Cloud, you'll be prompted to do so (using your GitHub account to sign in). Click `GET LINK` to create the snapshot. Click `OPEN` and Tilt will open the snapshot in Tilt Cloud in a new browser tab. Copy the URL of the snapshot and share at will.
 
+// TOOD: maybe a gif here instead?
 ![create-snapshot](assets/img/create-snapshot.png)
 
-## Viewing all private team snapshots
+Visit [cloud.tilt.dev/snapshots](https://cloud.tilt.dev/snapshots) for a list of all active snapshots you have created.
```
victorwuky, line 25:
```suggestion
Visit [cloud.tilt.dev/snapshots](https://cloud.tilt.dev/snapshots) for a list of all snapshots you have created.
```

Let's remove "active" since there's no notion of active vs inactive for snapshots.

docs/snapshots.md
```diff
@@ -1,44 +1,47 @@
...
 
-Click the `Create a snapshot` button in Tilt. In the modal, click `View snapshots from your team`. to see a list of all private team snapshots, for that team, in Tilt Cloud.
+### Private Snapshots
```
victorwuky, line 27:
```suggestion
### Private team snapshots
```

I think we should emphasize "team" here somehow. I want to make it explicit that there's only two types of snapshots:

- Team snapshots, which are private.
- Non-team snapshots, which are public.

I don't want a reader having to wonder, can I have a public team snapshot?

docs/snapshots.md
```diff
@@ -1,44 +1,47 @@
...
 
-## Public snapshots
+// TODO: screenshot (access denied)
 
-If your project is not associated with a Tilt Cloud team (i.e. your Tiltfile is not using `set_team`), you will still be able to create snapshots. Following the same flow as above, the snapshots you create will be public. Everyone on the internet will be able to access them, without having to sign into Tilt Cloud.
+**Important**: If the Tiltfile errors on its first execution when you run `tilt up`, and [`set_team`](https://docs.tilt.dev/api.html#api.set_team) has not been called yet, Tilt will not understand `set_team`. If you then take a snapshot, it will be a _public_ snapshot. (See below for public snapshots.) To work around this problem, make sure to put `set_team` as the first line in your Tiltfile. Tilt will then successfully parse `set_team` and start using the team id in it. Create a snapshot as normal and it will be a private team snapshot.
```
victorwuky, line 32:
```suggestion
**Important**: If the Tiltfile errors on its first execution when you run `tilt up`, and [`set_team`](https://docs.tilt.dev/api.html#api.set_team) has not been called yet, Tilt will not understand `set_team`. If you then take a snapshot, it will be a _public_ snapshot, not associated with any team. To work around this problem, make sure to put `set_team` as the first line in your Tiltfile. Tilt will then successfully parse `set_team` and start using the team id in it. Create a snapshot as normal and it will be a private team snapshot.
```

victorwuky:
> tangentially, I tried to make this doc less explicitly Laura-directed -- the points in here are largely meant to resonate with Laura, but I took out some of the explicit "you, as the DevEx head" language so that this doc can also serve as a resource for a Frieda being told to use snapshots who's wondering what the heck it is. However, I feel less strongly about this point than I do about the above point.

I don't have strong thoughts about this. But the existing latest doc is written by me and is explicitly Laura-directed, given what Dan B has said in the past about the feature. @dbentley : Please let us know if this is fine with you to make it less Laura-directed with Maia's changes.

docs/snapshots.md
maiamcc: Hmm I think that the phrasing "private team snapshots" does itself suggest the existence of "public team snapshots". What about:
> Private Snapshots (aka Team Snapshots)
?

victorwuky: Yep I like your latest suggestion

maiamcc:
@dbentley ping!

victorwuky:
@maiamcc : I would be comfortable if you decided to merge this (after fixing the TODOs). (I already approved last week.) I think the changes make sense and are generally good.

PR #676, opened 2020-07-16
docs: updated list of extensions with min_k8s_version()

abdullahzameek:
The TitlFile API Reference now has `min_k8s_version()` is now listed on the extensions section


abdullahzameek:
Merged into master!

PR #679, opened 2020-07-20
docs: if a user installed with install script, upgrade with install script

### LINKED ISSUE #643: Upgrade installs tilt to wrong path
https://github.com/tilt-dev/tilt.build/blob/master/docs/upgrade.md

Install script conditionally checks for .local:

```
 if [[ ":$PATH:" == *":$HOME/.local/bin:"* ]]; then
      mv tilt "$HOME/.local/bin/tilt"
```

However upgrade instructions simply move the binary to /usr/local:

```
curl -L https://github.com/tilt-dev/tilt/releases/download/v0.14.3/tilt.0.14.3.mac.x86_64.tar.gz | tar -xzv tilt && \
  sudo mv tilt /usr/local/bin/tilt
```

So upgrading won't affect the version of tilt you have installed if tilt is installed in `.local`.

Suggest that upgrade instructions should simply encourage user to rerun install script?

jazzdan:
Fixes #643

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -18,15 +18,33 @@ $ brew update && brew upgrade tilt-dev/tap/tilt
 $ brew install tilt-dev/tap/tilt
 ```
 
-### If you installed with the install script or directly with a release binary
+### If you installed with the install script
+
+Rerun the install script:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh | bash
+```
+
+### If you installed directly with a release binary
 
 ```bash
 curl -L https://github.com/tilt-dev/tilt/releases/download/v0.16.1/tilt.0.16.1.mac.x86_64.tar.gz | tar -xzv tilt && \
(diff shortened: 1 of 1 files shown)


docs/upgrade.md
```diff
@@ -18,16 +18,34 @@ $ brew update && brew upgrade tilt-dev/tap/tilt
 $ brew install tilt-dev/tap/tilt
 ```
 
-### If you installed with the install script or directly with a release binary
+### If you installed with the install script
```
nicks, line 21: note that the install script uses homebrew, so it feels weird to put homebrew above the install script in the list of options

can we reorganize this to be more like the install page?

https://docs.tilt.dev/install.html

i.e., point people to the installer script as the primary upgrade path, and give them alternative options at the bottom?

note that this is also missing the windows install script (which is another point in favor of reorganizing, since we seem to be missing things)

jazzdan: Yeah I like that idea, will do

jazzdan: @nicks reorganized! LMK what you think

PR #683, opened 2020-07-23
docs: more documentation on deploy tags and why we use them

nicks:
Hello @jazzdan, @maiamcc,

Please review the following commits I made in branch nicks/ch8524:

75c9c7baa89fb527a5975232ff52cf29ee5028f3 (2020-07-23 18:37:09 -0400)
docs: more documentation on deploy tags and why we use them

### DIFF
--- a/docs/custom_build.md
+++ b/docs/custom_build.md
@@ -102,22 +102,59 @@ There are a couple of caveats you should be aware of with `buildah` (and similar
 
 All the commands above contain `$EXPECTED_REF`. What is that?
 
-### Tags
+Tilt always pushes a content-based, immutable tag, not a bare ref. (Instead of `gcr.io/company-name/frontend`, Tilt injects `gcr.io/company-name/frontend:tilt-ffd9c2013b5bf5d4`, where the `ffd9c2013b5bf5d4` part is based on the contents of your image). Before explaining why (see [below](#why-tilt-uses-immutable-tags)), let's describe what this means for your Tiltfile and build script.
 
-Tilt always deploys with a digest, not a bare ref. (Instead of `gcr.io/company-name/frontend`, Tilt injects `gcr.io/company-name/frontend:tilt-ffd9c2013b5bf5d4`). Before explaining why (at the bottom of this document), let's describe what this means for your Tiltfile and build script.
+There are two ways for Tilt and your build script to coordinate image builds.
 
-Because different build tools have different ergonomics, Tilt supports two modes:
-* one-time tags via $EXPECTED_REF
-* temporary refs
+### The Good Way
 
-### One-time Tags
-This mode is easy if your tool takes the destination of the image as an argument (e.g. `docker build`).
-* Tiltfile sets a custom build command (e.g. `custom_build(..., 'docker build -t $EXPECTED_REF frontend')`).
-* Tilt sets a one-time ref as environment variable before executing the custom build command. (e.g. `EXPECTED_REF=gcr.io/company-name/frontend:tilt-ffd9c2013b5bf5d4`)
-* The custom build command builds the image and pushes to that ref (e.g. by reading `$EXPECTED_REF`).
+Most tools take a destination of the image as an argument (e.g. `docker build`).
 
-### Temporary Refs
-Other tools want to have an image ref hard-coded in configuration. They'll build and push to the same place each time. Instead of having to change your tool, tell Tilt what tag the build image will have with `custom_build(..., tag='frontend:s2i')`. After Tilt runs your build command, it will find this image and retag and push it with a unique ID.
+* Before running your build script, Tilt sets the environment variable
+  `$EXPECTED_REF` with a randomized tag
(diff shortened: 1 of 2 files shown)


nicks:
documentation described as https://github.com/tilt-dev/tilt/issues/3584

I also rewrote the bit about tags in custom_build to match what I've seen in other docs/presentations that I've seen in the kubernetes community

docs/custom_build.md
```diff
@@ -102,23 +102,44 @@ There are a couple of caveats you should be aware of with `buildah` (and similar
 
 All the commands above contain `$EXPECTED_REF`. What is that?
 
-### Tags
+### Build Tags
```
maiamcc, line 105: Imo you can hammer home the difference btwn build and deploy tags a bit, since i suspect ppl reading this won't initially realize they're different. consider either a section called "image tags" / "build and deploy tags" or two separate sections, one for "Build tags" and one for "deploy tags".

I also might explain "because different built tools..." first and then later differentiate: "regardless of the tag it BUILT with, Tilt always DEPLOYS with..."

docs/custom_build.md
```diff
@@ -102,23 +102,44 @@ There are a couple of caveats you should be aware of with `buildah` (and similar
...
 * The custom build command builds the image and pushes to that ref (e.g. by reading `$EXPECTED_REF`).
 
-### Temporary Refs
+### Temporary Build Refs
+
 Other tools want to have an image ref hard-coded in configuration. They'll build and push to the same place each time. Instead of having to change your tool, tell Tilt what tag the build image will have with `custom_build(..., tag='frontend:s2i')`. After Tilt runs your build command, it will find this image and retag and push it with a unique ID.
 
+### Deploy Tags
+
+In rare cases, another script in your build system may need to know what tag Tilt is going to deploy.
```
maiamcc, line 126:
```suggestion
In rare cases, another script in your build system may need to know what tag Tilt is going to _deploy_. Note that this tag is different from the one that Tilt will _build_ (which will be either provided in `$EXPECTED_REF` or hardcoded in the `tag` parameter, as explained above).

Tilt always deploys with a...
```

docs/custom_build.md
```diff
@@ -102,23 +102,44 @@ There are a couple of caveats you should be aware of with `buildah` (and similar
...
+
+Tilt has a special command to help with this. After you build the image, run:
```
maiamcc, line 128: IS there a comprehensible example you could give here? i think it would help me understand what cases this IS for so I wouldn't be tempted to use it in the wrong case

docs/custom_build.md
nicks: I ended up rewriting this whole section. I was trying to preserve as much as I could, but the more I read over it, the more I saw different things that were subtly wrong or confusing. Even the way this section calls them different types of refs is weird and inaccurate.

docs/custom_build.md
```diff
@@ -102,23 +102,44 @@ There are a couple of caveats you should be aware of with `buildah` (and similar
...
-Tilt always deploys with a digest, not a bare ref. (Instead of `gcr.io/company-name/frontend`, Tilt injects `gcr.io/company-name/frontend:tilt-ffd9c2013b5bf5d4`). Before explaining why (at the bottom of this document), let's describe what this means for your Tiltfile and build script.
+Tilt always deploys with a content-based, immutable tag, not a bare ref. (Instead of `gcr.io/company-name/frontend`, Tilt injects `gcr.io/company-name/frontend:tilt-ffd9c2013b5bf5d4`, where the `ffd9c2013b5bf5d4` part is based on the contents of your image). Before explaining why (see [below](#why-tilt-uses-immutable-tags)), let's describe what this means for your Tiltfile and build script.
 
-Because different build tools have different ergonomics, Tilt supports two modes:
+Because different build tools have different ergonomics, Tilt supports two modes for tagging an image in a custom_build:
 * one-time tags via $EXPECTED_REF
 * temporary refs
 
-### One-time Tags
+### One-time Build Tags
```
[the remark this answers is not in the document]
nicks: moot

docs/custom_build.md
nicks: moot

docs/custom_build.md
nicks: added a note at the top of the section

docs/custom_build.md
```diff
@@ -102,22 +102,59 @@ There are a couple of caveats you should be aware of with `buildah` (and similar
...
+
+Other tools have an image ref hard-coded in configuration. They'll build
+to the same tag each time.
+
+Instead of writing a wrapper script around your tool, tell Tilt what tag the
+build image will have with `custom_build(...,
+tag='gcr.io/company-name/frontend:dev')`.
+
+After Tilt runs your build command, it will find this image and retag and push
+it with a unique ID.
```
maiamcc, line 132:
```suggestion
it with a content-based tag.
```
to keep verbiage consistent

nicks: good catch! done

PR #684, opened 2020-07-24
Add Homebrew instructions to install docs

luizberti:
I noticed you folks [actively maintain a Homebrew tap](https://github.com/tilt-dev/homebrew-tap) for Tilt, but don't make any mention of it in the installation docs or the main repo's README.

Most macOS users that I'm aware of use Homebrew, and would prefer it to curling a script into `bash`. Defaulting to Homebrew is one less adoption barrier, and also means that the package can be upgraded much easier (`brew upgrade tilt`), and that it will auto-upgrade whenever users do a `brew upgrade`, helping you keep the Tilt userbase closer to the latest release :)

This PR, and [its accompanying PR](https://github.com/tilt-dev/tilt/pull/3613), add the missing documentation to make people aware of the `brew install` option

Also, there seems to have been an attempt to [upstream this into Homebrew's core tap](https://github.com/Homebrew/homebrew-core/pull/36309) that died out due to staleness. Maybe it would be worth trying to upstream it again?

Thanks for Tilt, it's been a really helpful tool so far 🎉

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -29,7 +29,13 @@ macOS
 kubectl config use-context docker-desktop
 ```
 
-- Install `tilt` with:
+- Install `tilt` with Homebrew:
+
+```bash
+brew install tilt-dev/homebrew-tap/tilt
+```
+
+- Alternatively without Homebrew, using the following command:
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh | bash


nicks:
Hi @luizberti ! Thanks for the PR!

I think you might have missed this part of the doc: https://docs.tilt.dev/install.html#homebrew-macos

We'd prefer to have a single installer script. That installer script checks if you have homebrew, and uses it if you do:
https://github.com/tilt-dev/tilt/blob/master/scripts/install.sh#L36
though it might help to make it clearer somewhere that this uses homebrew internally?

luizberti:
Ah, indeed I missed it!

That being said though, would you be open to revisiting that decision?

It really strikes me as an odd thing to have the _preferred_ installation method be curling a script into bash, and that was something that, personally, put me off from installing Tilt for ~1 week. Given that Kubernetes has a big target audience with picky sysadmins, I'd also imagine I'm not the only one who's not fond of this installation method, and who doesn't enjoy the proliferation of unmanaged packages installed to their computer.

Also, I do believe that it's in your best interest to do this, as there are several different ways this can harm adoption:
- Ideally your userbase should be subscribed to a PPA, RPM repo, or Homebrew Tap so that they can receive automatic updates, rather than having to re-run the installer to update. Having an updated userbase benefits you directly;
- Some users may have different opinions on how/where their packages should be installed, and can configure these details, as well as SELinux settings, etc by leveraging existing automation they have for package management, through `brew`/`apt`/`yum` directly or other tools that may build on top of these. Breaking away from this workflow makes tilt an "annoying" package to deal with;
- Some enterprise customers, due to internal policies or having to compliance with SOC2 or similar things, simply will not allow unmanaged packages to be installed, and will require that things get installed through the package manager, and only from authorized repos. This is because of various security implications related to having stale packages, or to the process of manual installs, especially through curling scripts into the interpreter, being insecure and error prone;

I think having it as a *possible* installation method is perfectly fine though, just maybe not as the default and recommended one. If I can by making a suggestion to help you with the package distribution situation, [Bazel](https://bazel.build) can do [pretty much all of the heavy lifting](https://github.com/bazelbuild/rules_pkg) related to building packages for different distros and architectures. Extending your Makefile would also be fine I suppose, although building debs and rpms can get finicky at times, many projects do that just fine.

And on that note, even if you decide to keep just the script for now, may I suggest merging my PRs anyways? The README currently makes no mention of Homebrew even though the Installation page on your docs does, but even on the docs, the reason I missed it when I first read it is that there is no mention of it under the `macOS` section, which is where one would expect to find this. I assume most people would also not imagine that there is another section at the bottom of the document for this.

jazzdan:
I think one reason we prefer the install script is because that handles OS detection. If we used the browser user agent to detect the OS, and showed you the appropriate screen, then it would be easier in my mind to show a more native installation method (homebrew, apt, etc) at the top.

nicks:
ya, i think your point is well-taken that we should say on the README that we support homebrew

i also take your point that we should have a PPA and/or RPM repo, and that most sysadmins will want to use a package manager instead of a one-off install script.

i do think there might be better ways we can visually present the different installers. We don't want to throw the user at a big decision tree of install options.

nicks:
researched a bit and filed https://github.com/tilt-dev/tilt/issues/3666

nicks:
The teams we work with seem happy with this approach, and can use the option tree at https://docs.tilt.dev/install.html if they need more. closing this pr

PR #687, opened 2020-07-28
make a schedule a demo modal popup

jazzdan:
Screenshot from 2020-07-28 16-26-13

Screenshot from 2020-07-28 16-26-32

Notes:

* The buttons are now slightly askew in the call to action. Paging @hyu for help :
* The close button should probably be styled better. Any ideas?
* The calendly widget isn't exactly centered inside the modal. Any ideas? Maybe it should be flexbox inside the modal?
* Any ideas for better copy?

### DIFF
--- a/src/_includes/index_cta.html
+++ b/src/_includes/index_cta.html
@@ -1,13 +1,13 @@
-<section class="popup-modal" data-popup-modal="one">
-  <h2 class="popup-modal__title">
+<section class="modal" data-modal="one">
+  <h2 class="modal__title">
     Schedule a Demo
   </h2>
-  <button class="popup-modal__close">{% svg assets/svg/x.svg %}</button>
-  <div class="popup-modal__content">
+  <button class="modal__close">{% svg assets/svg/x.svg %}</button>
+  <div class="modal__content">
     <div class="calendly-inline-widget" data-url="https://calendly.com/dbentley/tilt-enterprise?hide_event_type_details=1">
     </div>
-    <div class="popup-modal__content__text">
-      <p class="popup-modal__heroText">
+    <div class="modal__content__text">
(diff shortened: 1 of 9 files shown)


src/_includes/index_cta.html
```diff
@@ -1,10 +1,39 @@
+<section class="popup-modal" data-popup-modal="one">
```
hyu, line 1: I feel like "popup-modal" is redundant. Why not just "modal-" as the prefix?

src/_layouts/default.html
```diff
@@ -4,7 +4,7 @@
   {%- include head.html -%}
 
   <body class="body body--{{ page.layout }}">
-
+    <div class="body-blackout"></div>
```
hyu, line 7: I think this class should be something like "modal-background" to make it more obvious why it's here.

src/_sass/tilt.scss
```diff
@@ -132,6 +132,8 @@ $on-docs-mobile:    $header-width + $spacing-unit;
 
 // --------------------------------------------------
 
+$z-modal: 1500;
```
hyu, line 135: FYI, I used Sass vars for z-index so we can see this single list of how various elements are stacked up!

src/index.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Tilt
 layout: home
+has_calendly: true
```
hyu, line 4: FYI — Instead of having the Calendly JS inline, this bool in the header will inject it into the head of the HTML.

src/_layouts/default.html
hyu: Also, I really wish that clicking on this modal background would also dismiss the modal. It's annoying to snipe that "X" close button.

PR #689, opened 2020-07-30
docs: add notes on cluster detection

nicks:
Hello @maiamcc, @victorwuky,

Please review the following commits I made in branch nicks/clusterdetection:

c5c303776f1cd51ca1d2f35b428a8f5e591a6757 (2020-07-30 17:26:50 -0400)
docs: add notes on cluster detection

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -187,3 +187,26 @@ The most common options we see in local development are
 
 Check out our [Guide to Choosing a Local Cluster](choosing_clusters.html).
 
+---
+
+### Q: How does Tilt know which local Kubernetes cluster I'm using?
+
+To check which cluster Tilt is connecting to, run:
+
+```bash
+tilt doctor
+```
+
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -187,3 +187,26 @@ The most common options we see in local development are
...
+```
+
+The `Env` is the type of cluster Tilt has detected. Tilt uses this to determine
+which local dev features the cluster supports.
+
+In the dev cluster community, a common convention is to use the cluster name prefix
+to indicate the cluster type. For example,
+
+- Docker uses the `docker-` prefix
+- GKE uses the `gke_` prefix
```
victorwuky, line 207: @nicks : Is this supposed to be an underscore?

nicks: yep! this convention is ... not super consistent across implementations :grimacing:

PR #690, opened 2020-07-31
Remove dead pages: Product Direction, KubeCon

hyu:
The content is outdated and basically no one is looking at them.

## Hello (created for a prior KubeCon)
Screen Shot 2020-07-31 at 11 26 08 AM

## Product Direction
Screen Shot 2020-07-31 at 10 51 18 AM


nicks:
cc @victorwuky i think the product direction page was something Victor would point people to at customer meetings, tho true i don't think it's up-to-date anymore

victorwuky:
> cc @victorwuky i think the product direction page was something Victor would point people to at customer meetings, tho true i don't think it's up-to-date anymore

This is correct. I'd love to revive it at a future point. But certainly I have no problem in removing it now.

PR #706, opened 2020-08-11
site: Fix Schedule a demo button text so it is not clipped on mobile

victorwuky:
Could not reproduce it on Chrome/Safari on mac (using the developer tool console). But verified this on my physical iPhone 11. And debugged it using remote mobile web dev (i.e. connected phone to computer and ran the inspector on my computer against mobile safari running on phone).

| Before | After |
| --- | --- |
| IMG_3698 | IMG_3699 |

### DIFF
--- a/src/_sass/home.scss
+++ b/src/_sass/home.scss
@@ -111,9 +111,8 @@
   font-family: $tilt2-font-sans;
   font-size: $tilt2-button-font-size--large;
   font-weight: $tilt2-font-weight-sans-bold;
-  padding: 0;
   height: $tilt2-button-height--large;
-  width: 270px;
+  min-width: 270px;
   display: flex;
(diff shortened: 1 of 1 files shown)


src/_sass/home.scss
```diff
@@ -111,6 +111,7 @@
   font-family: $tilt2-font-sans;
   font-size: $tilt2-button-font-size--large;
   font-weight: $tilt2-font-weight-sans-bold;
+  padding: 0;
```
nicks, line 114: fwiw, it should probably be min-width instead of width and white-space: nowrap, so that the button never overflows (regardless of zoom level/font size)

victorwuky: Thanks @nicks . Fixed.

PR #712, opened 2020-08-13
Removing my profile from the About Page 

abdullahzameek:


src/_data/people.yml
```diff
@@ -42,8 +42,3 @@ victor:
   github: victorwuky
   twitter: victorwuky
   active: true
-abdullah:
-  name: Abdullah Zameek
-  github: abdullahzameek
-  twitter: abdullahzameek
-  active: true
```
nicks, line 49: actually, can you just change this to `false` and leave the user there?

that makes it easier for us to reference you in other places after you leave

nicks: (see how Maria's is done)

abdullahzameek: Ahhh I see, made that change!

PR #715, opened 2020-08-15
docs: add detail on dependencies and parallelism to local_resource docs

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/localresource:

22b2b868ed554215d144bd6c8884acb7d45d95f9 (2020-08-14 20:03:41 -0400)
docs: add detail on dependencies and parallelism to local_resource docs

### DIFF
--- a/docs/local_resource.md
+++ b/docs/local_resource.md
@@ -3,17 +3,18 @@ title: Run Local and/or Occasional Workflows with Local Resource
 description: "A technical reference on how to use local build steps in your Tiltfile"
 layout: docs
 ---
-(This is a technical doc; see the [Local Resource feature announcement blog post](https://blog.tilt.dev/2019/11/15/local-resource.html)
-for more context on this feature, and an explanation of some circumstances where it might come in handy.)
+Each entry in your Tilt sidebar is a **resource**---a unit of work managed by Tilt. 
 
-Each entry in your Tilt sidebar is a **resource**---a unit of work managed by Tilt. (For context,
-the most common type of Tilt resource is one that represents a deployed service, and is made up of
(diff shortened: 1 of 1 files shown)


docs/local_resource.md
```diff
@@ -43,6 +44,31 @@ but not in response to any particular file changes.
...
+your database comes up in Kubernetes.
+
+See the [Resource Dependencies guide](resource_dependencies.html) for more.
+
+## Parallelism
+
+Local resources can read and write to the local filesystem freely. 
+
+Tilt will run local resource in serial, to ensure that a file write from one resource doesn't
+interfere with another local resouce.
```
landism, line 61:
```suggestion
By default, local resources will not update at the same time as any other resources, to ensure that any files it writes don't cause problems. (e.g., infinitely re-triggering itself, or triggering another resource when its input is only half-ready).
```
1. worth clarifying that this only applies to update cmds, not run cmds
2. they're exclusive with all other resources, not just local resources
3. if we get more explicit about the problems this prevents, readers are more likely to be able to make a good decision about whether they want to use `allow_parallel`
  a. Maybe we should write more than what I put in parens in the suggestion.
  b. Maybe we should also say something like "if your local resource writes no files, this is always safe to enable".

nicks: hmmm...i don't think your new text is accurate. how would reducing parallelism stop infinite loops?

landism: > hmmm...i don't think your new text is accurate. how would reducing parallelism stop infinite loops?

Yeah, I think you're right. I'm not sure what case I was thinking of.

PR #727, opened 2020-09-03
Use experimental_analytics_report to send whoami name

victorwuky:


victorwuky:
I got this error when using `rstrip`: https://cloud.tilt.dev/snapshot/AZjoi_ULPECLpJH10uU=

I took out `rstrip` and it worked fine: https://cloud.tilt.dev/snapshot/Af7oi_ULYq_eT14RHrg=

landism:
> Update: I had to cast to a string to use rstrip

oy. yeah, I'd really like to improve the ergonomics of getting strings from locals

Tiltfile
```diff
@@ -2,6 +2,8 @@
 
 default_registry('gcr.io/windmill-public-containers')
 set_team('0584d8f6-05a2-49f5-923b-657afef098fe')
+username = str(local('whoami')).rstrip('\n')
+experimental_analytics_report({'user': username})
```
landism, line 6: There's an existing legacy tag named 'user', so we should probably avoid using that here.

How about 'user.name'?

victorwuky: Sure. Changed.

landism (approved):

PR #729, opened 2020-09-08
docs: document experimental_analytics_report

landism:
We only expect people to use this function in coordination with us. As such, it is documented on an unlinked page and is not listed in the main api docs. Account managers can choose to show or not show this page to customers, but this at least exists to give context to account managers.

### DIFF
--- a/docs/experimental_analytics_report.md
+++ b/docs/experimental_analytics_report.md
@@ -13,7 +13,7 @@ As the name says, this is *experimental*. It is intended only for use in coordin
 ## Usage
 `experimental_analytics_report` takes a single `Dict[string,string]` of tags to report per user. This function causes Tilt to report all data specified, plus the (anonymized) machine and repo hashes to the Tilt analytics server.
 
-Example:
+To work with existing dashboards, a Tiltfile needs to set the 'user.name' tag, e.g.:
 ```
 username = str(local('whoami')).rstrip('\n')
 experimental_analytics_report({'user.name': username})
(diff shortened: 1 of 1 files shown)


victorwuky (changes requested):
@landism : Can we please have stronger language to ask the audience reading this to use `user.name`? I definitely want to avoid creating custom queries for each customer because they used slightly different keys. (i.e. same chart for each customer, especially in light of : https://app.clubhouse.io/windmill/epic/9322/

docs/experimental_analytics_report.md
```diff
@@ -0,0 +1,27 @@
+---
+title: experimental_analytics_report
+description: "A Tiltfile builtin for better understanding your team's analytics"
+layout: docs
+---
+## Summary
+Tilt 0.17.4 adds the `experimental_analytics_report` Tiltfile builtin. This allows teams working with Tilt to better measure and improve their Tilt usage.
```
victorwuky, line 7:
```suggestion
[Tilt 0.17.4](https://github.com/tilt-dev/tilt/releases/tag/v0.17.4) adds the [`experimental_analytics_report` Tiltfile builtin](https://github.com/tilt-dev/tilt/pull/3751). This allows teams working with Tilt to better measure and improve their Tilt usage.
```

victorwuky (approved):

docs/experimental_analytics_report.md
landism: I had actually put links there to start, but on second thought figured that they didn't add much information that's relevant to users.

Hm, I think the link to Tilt 0.17.4 is fine, since that's a reasonable expectation when someone clicks on Tilt 0.17.4. I don't think linking to the PR is useful. As a user, I'd expect that to link to the API docs or some other kind of user-facing doc. AFAICT, the PR doesn't add any information and clicking on it would just be a disappointment and waste of the user's time. The PR feels like a "how the sausage is made" thing that I wouldn't expect to generally be important to our audience, and is especially uninteresting in this case. Do you have an angle I'm missing, or a different set of values we should align on?

victorwuky: @landism : Purely a non-blocking suggestion. Discard if you'd prefer.

PR #743, opened 2020-09-16
rollout: two more pages

dbentley:


victorwuky:
@dbentley : A common complaint about adoption/usage I hear a lot is that "adoption is just about people using it, it doesn't really say if people are really enjoying it or finding it valuable". In fact, one customer had a similar concern in a recent customer meeting.

I think the correct response to this, is that the "adoption metric" is good enough. And that it's really hard to measure perceived value directly. Adoption _does_ measure perceived value, but with a delay. But that's still good enough, because adoption is much easier to measure. Adoption is easy to measure because:
- Usually there's a metric to do so. The tooling is straightforward.
- But more importantly, an adoption metric is easy to understand and fairly objective. And so it's easy to use. 

Some of these points are already covered in https://docs.tilt.dev/rollout/active-hours. But I think it might be worth anticipating the concern of "adoption isn't good enough" as a metric, and have a one or two sentence explanation, and point to https://docs.tilt.dev/rollout/active-hours.

docs/rollout/why-could-this-be-fast.md
```diff
@@ -0,0 +1,40 @@
...
+## Breaking the Logjam
+
+Frustrated users often have a good basis for being frustrated, but that frustration can get in the way of communicating that reason. Asking "Why Could This Be Fast?" helps you both focus on potential improvements.
+
+For example, say a frustrated user complains "this build is taking 2 minutes". It's hard to know what to do with that: Scala users might be thrilled if their build only took 2 minutes. So you ask "why could it be faster?" When the user says "builds shouldn't have to download any dependencies when I don't change the set of dependencies", you go to the Dockerfile and improve caching.
+
+Eliciting the user's intuition illuminated how to fix it.
+
+
+## Stay with "Why", now "How"
```
victorwuky, line 28:
```suggestion
## Stay with "Why", not "How"
```

victorwuky:
@dbentley : Why did you use "could" instead of "should" in "Why could this be fast?" In [earlier iterations](https://docs.google.com/document/d/1Vk7DuV1A6X_YRhB0knM--5nEgWFTpTO-0aEAXYhn7JY/edit#heading=h.31cmpgc3mq4r) you used "should" and that makes more intuitive sense to me. Why "could"?

victorwuky (approved):

PR #745, opened 2020-09-16
docs: change the default cluster instructions to recommend Kind on Linux

nicks:
Hello @maiamcc,

Please review the following commits I made in branch nicks/ch9397:

5dde607b70ae6b2dbebeda0d491c08f9c09f1863 (2020-09-16 18:36:34 -0400)
docs: change the default cluster instructions to recommend Kind on Linux
Fixes https://github.com/tilt-dev/tilt/issues/3772

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -39,28 +39,17 @@ Linux
 - Install [Docker](https://docs.docker.com/install/)
 - Setup Docker as [a non-root user](https://docs.docker.com/install/linux/linux-postinstall/).
 - Install [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
-- Install [Microk8s](https://microk8s.io/):
-
-```bash
-sudo snap install microk8s --classic && \
-sudo microk8s.enable dns && \
-sudo microk8s.enable registry
-```
-
-- Make microk8s your local Kubernetes cluster:
-
-```bash
-sudo microk8s.kubectl config view --flatten > ~/.kube/microk8s-config && \
-KUBECONFIG=~/.kube/microk8s-config:~/.kube/config kubectl config view --flatten > ~/.kube/temp-config && \
-mv ~/.kube/temp-config ~/.kube/config && \
-kubectl config use-context microk8s
-```
-
+- Install [Kind with a a local registry](https://github.com/tilt-dev/kind-local/blob/master/README.md)
 - Install `tilt` with:
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh | bash
 ```
+
+Alternatively, Ubuntu users sometimes prefer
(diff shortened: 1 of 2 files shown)


nicks:
i'm not totally sure this is the right recommendation. i personally still use microk8s every day as my go-to cluster. But I understand the complaints about microk8s, particularly on non-ubuntu platforms.

docs/install.md
```diff
@@ -39,29 +39,15 @@ Linux
...
-```
-
+- Install [Kind with a a local registry](https://github.com/tilt-dev/kind-local/blob/master/README.md)
 - Install `tilt` with:
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh | bash
 ```
 
+Alternatively, Ubuntu users sometimes prefer [Microk8s](choosing_clusters.html#microk8s) instead of Kind.
```
maiamcc, line 49: is it worth it to note here WHY you might go for microk8s over Kind?

maiamcc (approved):
ulllp! well, worth a shot!

maiamcc:
looks like we're still recommending D4D for MacOS, how do you feel about changing that rec to be Kind as well? I think it could also be good to put a tl/dr "our rec for your OS" at the top.

lmk if you want me to do the above, or if you want to do it, either is fine with me

docs/install.md
nicks: done

nicks:
re: D4D - this is a way outside the scope of the original user complaint linked above, but I'll take the bait:

Different docs have different audiences.

In the install guide, the audience is trying to figure out how to get Tilt setup as quickly as possible. For that audience, I think D4D is 100% the right recommendation.

But the audience of a 10 minute video about how to choose a local dev cluster is a different audience, and might be a different/more nuanced recommendation.

docs/install.md
```diff
@@ -39,29 +39,18 @@ Linux
...
-- Make microk8s your local Kubernetes cluster:
-
-```bash
-sudo microk8s.kubectl config view --flatten > ~/.kube/microk8s-config && \
-KUBECONFIG=~/.kube/microk8s-config:~/.kube/config kubectl config view --flatten > ~/.kube/temp-config && \
-mv ~/.kube/temp-config ~/.kube/config && \
-kubectl config use-context microk8s
-```
-
+- Install [Kind with a a local registry](https://github.com/tilt-dev/kind-local/blob/master/README.md)
```
ellenkorbes, line 42: > Kind with a a local registry

Typo!

ellenkorbes:
Also might be worth noting for the Windows install that those commands should be typed on Powershell, not on CMD.

docs/install.md
nicks: :grimacing: https://github.com/tilt-dev/tilt.build/pull/748

PR #747, opened 2020-09-18
rollout: flesh out/clean up Introduce Tilt doc

maiamcc:
Re-org and flesh out the Introduce Tilt doc.

This started as a copy-edit, I ended up making more changes here than anticipated whoops, but i think it adds a lot of useful context / squeezes in some other docs tips we've discussed in the past.

Also cleaned up a lot of typos/clarity stuff

### DIFF
--- a/docs/rollout/introduce-tilt.md
+++ b/docs/rollout/introduce-tilt.md
@@ -1,63 +1,81 @@
 ---
-title: Tilt Transition Guide
+title: Introducing Tilt with a Tilt Transition Guide
 layout: rollout
 ---
+You can have the greatest tool in the world, and it won't mean a thing unless you can show AppDevs how to use it and persuade them that they should care. This article will walk you through our recommended structure for a "Transition Guide", which is a document intended to help AppDevs transition off of their current workflow and to a new tool (in this case, Tilt, but the format can be generalized to any tool).
 
-Write a transition guide that will make an AppDev excited to try Tilt and able to see benefit in minutes.
+## What's a "Transition Guide"?
 
-* you'll like Tilt more because...
-* how to get started
-  * what to do after you start Tilt
-* advanced features
-* known issues
-* what to do next
+You've probably read many a README for an internal tool, but a transition guide is more than just a README: it's your case to an AppDev that this _new_ workflow is better than their _old_ workflow.
 
-If you have existing docs, it's worth tailoring it to this audience. For example, a design doc might cover much of the same material but it's meant to discuss implementation trade-offs. The goal of a transition guide is that a single link will let an AppDev get started.
+AppDevs have generally sunk lots of time and effort into their tooling. In our experience, it takes an especially persuasive argument to get an AppDev to switch from a tried-and-true workflow to something new. They need to know not just why the new tool is _good_, but why it's _better than what they're currently using_. 
+
+A good Transition Guide should illustrate the benefits of the new tool vis-a-vis the old tool, and give clear enough setup instructions that an AppDev can experience those benefits within minutes.
+
+Here's our recommended format (we'll dig into each section below):
+
+```
+1. You'll Like Tilt More Because...
+2. How to Get Started
+  2a. How to Install Tilt
+  2b. What to Do After You Start Tilt
+4. Known Issues
+5. What to Do Next
+```
+
+Remember that a Transition Guide is more than just a README. It's job is to get an AppDev up and running with minimal friction, but also to drive home why they should be excited about the new tool.
+
+It's also not a design doc or implementation proposal: focus on what makes this tool compelling to a day-to-day developer, not to a DevEx team (e.g. "easy to maintain") or a VP of Eng (e.g. "will boost overall productivity by X%").
 
 
-## You'll Like Tilt More Because...
+### You'll Like Tilt More Because...
 
-Start with why Tilt is better, and focus on why it's better for the AppDev right now. A piece of perspective: some Tilt benefits are more exciting to the DevEx who have to support the tooling than to the AppDevs using it. Instead of leading with attributes like "easier to maintain" or "closer to prod", we'd suggest focusing on advantages like being faster, having a better UI, or supporting new workflows.
+Start with why Tilt is better than the current way of doing things. Don't just list features of Tilt: highlight Tilt's benefits as compared to the legacy tool.
 
-Highlight benefits of Tilt for your project compared to the legacy tool. Don't just list features of Tilt: some may not be new compared to the legacy tool and some you may not use yet.
+This part of the doc can also be a great place to re-contextualize or justify what AppDevs might see as sticking points. For instance, if initial `tilt up` is noticeably slower than launching your old tool, consider a note like:
+> You may notice that Tilt is a little slower to start up than $OLDTOOL--but if you stick with it, you'll see that once your app is up and running, _updates_ are 10x faster than on $OLDTOOL.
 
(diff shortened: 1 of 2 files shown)


src/_data/rollout.yml
```diff
@@ -14,14 +14,14 @@
   - title: Non-Server Workflows
 - title: Onboard
   items:
-  - title: Introduce Tilt
-    href: rollout/introduce-tilt
   - title: Focus Users
```
maiamcc, line 17: Put "focus users" first because it seems like chronologically the first step

docs/rollout/introduce-tilt.md
```diff
@@ -1,63 +1,80 @@
...
+This part of the doc can also be a great place to re-contextualize or justify what AppDevs might see as sticking points. For instance, if initial `tilt up` is noticeably slower than launching your old tool, consider a note like:
+> You may notice that Tilt is a little slower to start up than $OLDTOOL, but if you stick with it, you'll see that once your app is up and running, _updates_ are 10x faster than on $OLDTOOL.
+
+A Transition Guide is about making your case to an AppDev, so focus on the features that most benefit someone who writes code day-to-day. Instead of leading with attributes like "easier to support users" or "dev is closer to prod." (which are more exciting to those maintaining a tool than those using it), we'd suggest focusing on advantages like being faster, having a better UI, or supporting new workflows.
+
+Remember that screenshots/screen recordings can be great tools to get your point across, and may be particularly good for comparing Tilt to the legacy tool. (If your team uses Tilt Cloud, you can also use [Snapshots](../snapshots) to let potential users interact with the UI before installing Tilt.)
```
maiamcc, line 39: This is actually a lousy experience currently, should we hold off on suggesting it until we have a better "give new TC users access to snapshots" flow?

docs/rollout/introduce-tilt.md
```diff
@@ -1,63 +1,80 @@
...
+1. You'll Like Tilt More Because...
+2. How to Get Started
+  2a. How to Install Tilt
+  2b. What to Do After You Start Tilt
+3. Known Issues
+4. What to Do Next
+```
 
-Highlight benefits of Tilt for your project compared to the legacy tool. Don't just list features of Tilt: some may not be new compared to the legacy tool and some you may not use yet.
+Remember that a Transition Guide is more than just a README. Its job is to get an AppDev up and running with minimal friction, but also to drive home why they should be excited about the new tool.
```
landism, line 26:
```suggestion
Remember that a Transition Guide is more than just a README. Its job is to get an AppDev up and running with minimal friction, but also to drive home why they should be excited about Tilt.
```

docs/rollout/introduce-tilt.md
```diff
@@ -1,63 +1,80 @@
...
+Start with why Tilt is better than the current way of doing things. Don't just list features of Tilt: highlight Tilt's benefits as compared to the legacy tool.
 
-### What To Do After You Start Tilt
-It's useful after you say how to start Tilt to give a simple example workflow that shows off getting quick feedback with Tilt. For example:
+This part of the doc can also be a great place to re-contextualize or justify what AppDevs might see as sticking points. For instance, if initial `tilt up` is noticeably slower than launching your old tool, consider a note like:
+> You may notice that Tilt is a little slower to start up than $OLDTOOL, but if you stick with it, you'll see that once your app is up and running, _updates_ are 10x faster than on $OLDTOOL.
+
+A Transition Guide is about making your case to an AppDev, so focus on the features that most benefit someone who writes code day-to-day. Instead of leading with attributes like "easier to support users" or "dev is closer to prod." (which are more exciting to those maintaining a tool than those using it), we'd suggest focusing on advantages like being faster, having a better UI, or supporting new workflows.
```
landism, line 37: Maybe something like "if you have focus users (test pilots?) (with a link), consider asking them how Tilt is particularly useful for your team"

landism: ah, I see you mentioned something like this down under "advanced features"...I feel like it's better here, or maybe fine in both, but then should probably adjust the wording of the second instance

docs/rollout/introduce-tilt.md
```diff
@@ -1,63 +1,80 @@
...
 
-At the beginning there are probably ways that Tilt is worse or even just different than the legacy dev environment. Telling users about these differences before they experience them builds trust. Soemthing as simple as a list of links to issues in your issue tracker helps manage users' expectations.
+### Next Steps
 
-## Next Steps
+Tell users how to give feedback and get support (by joining a Slack channel, subscribing to an email list, etc.). If you're looking for volunteers to help improve the Tilt experience, mention that here.
 
-Tell users how to give feedback and get support. Mabye they should join a Slack channel or subscribe to an email list. If you're looking for volunteers to help improve Tilt, mention that here.
+Your goal in this section should be to catch underwhelmed users before they fall out of the funnel---it's better to get their feedback now than to hunt them down later as [inactive users](../rollout/prioritize-inactive). The easier (or more mandatory) you make it for AppDevs to give you feedback, the greater the chance an underwhelmed AppDev will come give you useful information (rather than giving up and walking away).
```
landism, line 64: > or more mandatory

...mandatory feedback? this sounds weird

docs/rollout/introduce-tilt.md
```diff
@@ -1,63 +1,80 @@
...
 
-As you configured Tilt you may have added convenient features in the Tiltfile. E.g. making it easy to recreate the database with a [`local_resource`](../local_resource) or use config values to choose what servers to set up as editable. Tell users about the bells and whistles so they can enjoy them too!
+### Known Issues
 
-## Known Issues
+As you first build out Tilt support, there will probably be ways that Tilt is worse than the legacy tool---or even just different. Telling users about these regressions before they run into them builds trust. Something as simple as a list of links to issues in your issue tracker helps manage users' expectations. (As noted above, many perceived "regressions" stem from misconceptions about how Tilt works, or are necessary trade-offs for big benefits down the road: explaining this context can help your users feel less annoyed.)
 
-At the beginning there are probably ways that Tilt is worse or even just different than the legacy dev environment. Telling users about these differences before they experience them builds trust. Soemthing as simple as a list of links to issues in your issue tracker helps manage users' expectations.
+### Next Steps
```
landism, line 60: This is called "What to Do Next" in the outline above. They should match. I'm not sure which I prefer. I actually think neither is a good description of this section.
I haven't come up with a name I'm super happy with, but some spitballing:
* Suggestion box
* Let us know what you think
* How to get help
* "This sucks--who do I complain to?"

docs/rollout/introduce-tilt.md
landism, line 64:
```suggestion
Your goal in this section should be to catch underwhelmed users before they give up entirely---it's better to get their feedback now than to hunt them down later as [inactive users](../rollout/prioritize-inactive). The easier (or more mandatory) you make it for AppDevs to give you feedback, the greater the chance an underwhelmed AppDev will come give you useful information (rather than giving up and walking away).
```

It seems better to drop "funnel" to 1) not burden readers with thinking about that and 2) sound less growth-hacky

landism (approved):

docs/rollout/introduce-tilt.md
maiamcc: good call

maiamcc: 😬

docs/rollout/introduce-tilt.md
maiamcc: good call

maiamcc:
@dbentley @victorwuky sorry to add you late, apparently it didn't go through the first time. merging this in now but happy for further feedback

PR #754, opened 2020-09-23
Finish the POC

victorwuky:


docs/rollout/finish-poc.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Finish the POC
+layout: rollout
+---
+
+## Finish the POC
+
+As a DevEx engineer, do invest the time in exploring Tilt, possibly creating a POC (Proof of Concept) implementation of the Tiltfile, and getting feedback from a select few AppDevs and other colleagues. But be wary of extending the POC phase too long. Set a clear goal to finish the POC by a given date, so that you can really focus on starting to roll out Tilt to AppDevs, supporting their daily use in a "production" sense.
```
dbentley, line 8: I think date is okay; it might be even better to say "what are we trying to prove with the POC" and prove those, to know when it's done.

dbentley (approved):

docs/rollout/finish-poc.md
victorwuky: Yes great point @dbentley . I vaguely remember my brain thinking this, but was too lazy to latch onto this passing thought. I've updated so it's not time-based, but outcome based.

PR #761, opened 2020-09-30
api: preliminary docs for url/port forward changes (won't merge until code work done) [ch9604]

maiamcc:
Docs stubbing out proposed Tiltfile API for improvements to resource url discoverability/accessibilty. Feedback/comments welcome. Won't merge these til features are implemented, but if I get a LGTM I'll start implementing!


api/api.py
```diff
@@ -20,6 +20,35 @@ class LiveUpdateStep:
...
+
+  By default, the host for a port-forward is ``localhost``. This can be changed with
+  the ``--host`` flag when invoking Tilt via the CLI.
+
+  Attributes:
+    local_port (int): the local port to forward traffic to.
+    container_port (int, optional): if provided, the container port to foward traffic _from_. If not provided, Tilt will forward traffic from ``local_port``, if exposed, and otherwise, from the first default container port. E.g.: ``PortForward(1111)`` forwards traffic from container port 1111 (if exposed; otherwise first default container port) to ``localhost:1111``.
+    link_text (str, optional): if provided, the text of this URL when displayed in the Web UI. This parameter can be useful for disambiguating between multiple port-forwards on a single resource, e.g. labeling one link "App" and one "Debugger." If not given, the Web UI displays the URL itself (e.g. "localhost:8888").
+  """
+  def __init__(self, local_port: int,
```
nicks, line 35: the other classes in our api tend to have a function (e.g., `blob()`, `run()`, `sync()`) rather than a class constructor (which has slightly different semantics). is there a reason why you chose a different approach here?

api/api.py
```diff
@@ -20,6 +20,35 @@ class LiveUpdateStep:
...
+  """
+  Specifications for running and displaying a Kubernetes port-forward.
+
+  By default, the host for a port-forward is ``localhost``. This can be changed with
+  the ``--host`` flag when invoking Tilt via the CLI.
+
+  Attributes:
+    local_port (int): the local port to forward traffic to.
+    container_port (int, optional): if provided, the container port to foward traffic _from_. If not provided, Tilt will forward traffic from ``local_port``, if exposed, and otherwise, from the first default container port. E.g.: ``PortForward(1111)`` forwards traffic from container port 1111 (if exposed; otherwise first default container port) to ``localhost:1111``.
+    link_text (str, optional): if provided, the text of this URL when displayed in the Web UI. This parameter can be useful for disambiguating between multiple port-forwards on a single resource, e.g. labeling one link "App" and one "Debugger." If not given, the Web UI displays the URL itself (e.g. "localhost:8888").
```
nicks, line 33: wonder if we should call this `name` or use some other existing terminology in the kubernetes ecosystem

api/api.py
```diff
@@ -20,6 +20,35 @@ class LiveUpdateStep:
...
+               container_port: Optional[int] = None,
+               link_text: Optional[str] = None):
+    pass
+
+class Link:
+  """
+  Specifications for a link associated with a resource in the Web UI.
+
+  Attributes:
+    url (str): the URL to link to
```
nicks, line 45: seems like this should be a path rather than a url? since we don't want the user to hard-code the host/scheme here

api/api.py
maiamcc: oh! none in particular, I can use a constructor function! (what are the semantic differences of note, out of curiosity?)

api/api.py
maiamcc: This is for attaching an arbitrary URL (NOT a port-forward URL) which might be on the internet/localhost/some other host--I imagined that we DID want users to hard-code this rather than trying to provide all the options for assembly. Happy to be wrong about this tho. See: https://github.com/tilt-dev/tilt/issues/3571

I suppose we could have params for `scheme`, `host`, and `path` (and maybe even allow user to specify those OR `url`)? I honestly don't have a good sense of the sorts of URLs ppl will be using this to attach, and how frequently they'll be self-hosted vs. on the internet.

nicks: hmmm...ok, can you write out a (very short) spec of the different options here?

e.g.,

Suppose you have:

```
k8s_resource('fe', port_forwards='8080', links=link('/home', 'home'))
```

tilt up --host=127.0.0.1

Possible options:
- Tilt emits an error because '/home' is missing a scheme/port
- Tilt displays a link to localhost:8080/home
- Tilt displays a link to 127.0.0.1:8080/home

I could also imagine other interesting case studies for, e.g., if you have multiple port-forwards

I think this is something you could very quickly hallway test with L or Han and see what they think should happen. But we should have them written down there, since i think there's a lot of behavior here that could be super ambiguous

maiamcc: Ah, I understand the confusion: the `links` arg / `Link` type are currently only for use with `local_resource` to attach a complete URL. for `k8s_resource`, you should use a PortForward to set the port(s) and optionally text for the link. Host is still defined with `--host`. I'm calling "specifying a path on a port forward" out of scope for now but it should be easy enough to add that with an optional `path` arg on `PortForward`. Does that make sense? Or do you still think that attaching arbitrary URLs to a _local resource_ should support `host`?

nicks: ah, ok

api/api.py
nicks: the capital vs lowercase is the most salient (but important!) one. 

i believe that in core python, there are some differences in what operations they support

starlark has no built-in classes, so you'd have to simulate it with structs

nicks (approved):

PR #764, opened 2020-10-08
Clarify remote in job outro

victorwuky:


victorwuky:
@nicks : Does it make sense to remove this? I don't think there's any clear plans to go back to an office in NYC anytime soon right? And 4 of us aren't in NYC right now anyways right?

nicks:
can you explain your concerns a bit more? I do several phone screens a week, and the geographic distribution of the team is a people ask about / want to talk about, so i'd be in favor of adding more info rather than less

victorwuky:
@nicks : Sure. I just find the "Remote is OK" thing a bit weird, especially in pandemic times when everyone is doing remote. I'm worried it would be turning away candidates that think we don't really like remote or think it's a second-class citizen. If we just removed "Remote is OK" and leave the remark about NYC, I'm fine with that too. (Since the line below gives the geographical constraint already.)

nicks:
How about something like:

"About half our team is in NYC and working from home during the pandemic. About half is remote, each in a different city. We believe you can be successful in this role remotely."

fwiw, we've had plenty of strong remote candidates apply, so i'm not super concerned about that, more concerned that we're upfront about the relevant information

victorwuky:
Sounds good to me @nicks . Updated PR with your suggestion.

nicks (approved):

PR #775, opened 2020-10-28
docs: add api docs for load_dynamic

nicks:
Hello @landism,

Please review the following commits I made in branch nicks/load-dynamically:

24fb78ca0505acaa02a9d4af7085e8b490c1b389 (2020-10-28 17:57:00 -0400)
docs: add api docs for load_dynamic

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -442,12 +442,9 @@ def filter_yaml(yaml: Union[str, List[str], Blob], labels: dict=None, name: str=
   pass
 
 def include(path: str):
-  """Include another Tiltfile.
+  """Execute another Tiltfile.
 
-  Loads any builds or resources defined in that Tiltfile.
-
-  If you want to define common functions or constants and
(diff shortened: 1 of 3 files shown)


api/api.py
```diff
@@ -468,12 +463,41 @@ def load(path: str, *args):
...
+  Used when you want to define common functions or constants
+  to share across Tiltfiles.
+
+  Example ::
+
+    symbols = load_dynamic('./lib/Tiltfile')
+    create_namespace = symbols['create_namespace']
+    create_namespace('frontend')
+
+  Like :meth:`load`, each Tiltfile can only be executed once. Can also be used to load a `Tilt Extension <extensions.html>`_.
```
landism, line 495:
```suggestion
  Like :meth:`load`, each Tiltfile will only be executed once. Can also be used to load a `Tilt Extension <extensions.html>`_.
```

nit - "can" makes me think that a repeated load is an error

landism (approved):

PR #780, opened 2020-11-06
docs: Custom port forward names and additional links for resources

victorwuky:
screencapture-localhost-4001-accessing-resource-localhost-html-2020-11-09-14_54_14


docs/custom_port_forwards_and_links.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Custom port forward names and additional links for resources
```
nicks, line 2: This feels extremely narrow, and assumes a lot of context that the user reading it probably doesn't have. Like, I'm not sure that most users even understand what a port-forward is if they haven't used kubernetes before.

Can we make this more user-centric?

e.g.,

title: Exposing your service on localhost

outline:

- Tilt lets you expose your service on a port on localhost. We call these port-forwards!
- Here's what one port-forward looks like
- Let's give it a name
- Here's what two port-forwards looks like

nicks: or maybe just "How to find your server"

victorwuky: Thanks @nicks . Your comments are interesting because it reflects what I was thinking of in the context of https://app.clubhouse.io/windmill/story/10082/add-additional-docs-for-https-app-clubhouse-io-windmill-epic-6144-resource-url-discovery-accessibility and Maia's comments. 

The API docs have "narrow" explanations and what each API function does. And that's what I was going for in this addition. But I also see all the existing docs that are outside of the API page are more elaborate and tutorial-like and provide more context.

I'm happy to make this doc more in that direction following your outline suggestion.

nicks: ya, i don't think it's a big change, it's just re-contextualizing the doc a bit to focus more on what the user is trying to accomplish, rather than focusing on the API reference

I also think you could put it up in the "Onboarding your team" section, since we see this as pretty core tilt functionality

maiamcc: Yeah I agree with nick, and might pull out some of the explanation of Port Forwarding from Tiltfile Concepts (which is a doc that got majorly kitchen-sink'd) and put it here instead. imo this doc will be more useful if it also gives context about port forwards/endpoints.

docs/custom_port_forwards_and_links.md
```diff
@@ -0,0 +1,55 @@
...
+layout: docs
+---
+
+By default, port forward links for resources are displayed as `localhost:<port-number>` in Tilt, at the top of the screen. This may be confusing if you have multiple port forwards per resource, such as:
+
+<figure>
+    <img src="/assets/img/port-forwards.png" title="Port forwards">
```
maiamcc, line 9: consider at least one image that shows more of the UI (with maybe the URLs circled/highlighted?) so user can orient themselves in the full UI. Right now this image gives no clue as to where in the UI it happens.

victorwuky:
@maiamcc @nicks : Can you please review again? Incorporated your respective feedback.

> Yeah I agree with nick, and might pull out some of the explanation of Port Forwarding from Tiltfile Concepts (which is a doc that got majorly kitchen-sink'd) and put it here instead. imo this doc will be more useful if it also gives context about port forwards/endpoints.

Maia: I didn't remove anything from Tiltfile Concepts, but I did take content from that page. I was thinking we could adjust that page after this PR, if at all.

docs/accessing_resource_localhost.md
```diff
@@ -0,0 +1,91 @@
+---
+title: Accessing your resource on localhost
```
maiamcc, line 2: If you're using custom links, they're likely not on localhost. maybe "Accessing your resource" or "Surfacing links to your resource" or "Surfacing resource endpoints" or something

docs/accessing_resource_localhost.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Access your resource with Kubernetes port forwarding
+
+As you're configuring [resources in Tilt](/tiltfile_concepts.html#resources), you'll often want to access an app within a Kubernetes resource, for example, a web app or a database server. Kubernetes provides [port forwarding](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/) to allow you to access via your local machine with `localhost`. Tilt exposes this feature through a configuration in [`k8s_resource`](/api.html#api.k8s_resource). For example, if you have in your Tiltfile: 
+
+```
+k8s_resource(workload='blog-site', port_forwards=4002)
+```
+
+`localhost:4002` will be connected to container port 4002. Tilt also displays a link at the top of the screen too.
```
maiamcc, line 14:
```suggestion
`localhost:4002` will be connected to container port 4002. Tilt also displays a link at the top of the screen.
```

docs/accessing_resource_localhost.md
```diff
@@ -0,0 +1,91 @@
...
+
+With:
+
+```
+k8s_resource(workload='blog-site', port_forwards='9000:4002')
+```
+
+`localhost:9000` will be connected to container port 4002.
+
+## Multiple port forwards
```
maiamcc, line 28: FFTI: I might call this section `### Naming Port Forwards` instead, because naming PF's can be useful even if you don't have more than one (and then after you introduce the syntax, have a line that says "this can be especially useful if you have multiple port forwards and want to disambiguate between them" or something).

docs/accessing_resource_localhost.md
```diff
@@ -0,0 +1,91 @@
...
+)
+```
+
+would show this in Tilt:
+
+<figure>
+    <img src="/assets/img/additional-resource-links.png" title="Additional resource links">
+</figure>
+
+Arbitrary links applies to [`local_resource`](/api.html#api.local_resource) also, not just `k8s_resource`.
```
maiamcc, line 91:
```suggestion
Arbitrary links are supported on both [`local_resource`](/api.html#api.local_resource) and `k8s_resource`.
```

maiamcc (approved):
Are you going to talk about the `path` param at all?

victorwuky:
> Are you going to talk about the link_path param at all? (for specifying the path on the endpoint to link to--e.g. maybe there's nothing at localhost:8000, the app lives at localhost:8000/v1/app.)

> you might also mention how the --host flag interactions with port forwards, or at least link to the CLI docs.

@maiamcc : Good ideas. I can work on these later. But don't want to hold up this update. (And want to share with customers soon.)

docs/accessing_resource_localhost.md
victorwuky: Good call. Changed.

docs/accessing_resource_localhost.md
victorwuky: Good idea. Updated the heading. (I already talk about the disambiguating use case in the text.)

PR #787, opened 2020-11-13
add 'tilt demystified' to blog and docs

ellenkorbes:
@nick Gonna need a hand with two things here: One is that the thumbnail in the front page of the blog looks too big, not sure why. And the other is that thing I mentioned about the video embed. On the docs side things look okay as is IMO.


nicks:
cool looking at it now! i assume it's ok if i push changes to the branch?

ellenkorbes:
@nicks Yeah, of course :)

nicks:
cool, i fixed the video and compressed the images a bunch

the bigger thumbnail on the homepage is OK! That's how it's supposed to look! https://blog.tilt.dev/

nicks (approved):
Looks great!

PR #788, opened 2020-11-17
fix typo in `custom_build` args `ignore` key

friend0:
On Tilt v0.17.11, as documented, the `ignores` key will throw an error
Using `ignore` is accepted


nicks (approved):
thanks! LGTM

PR #796, opened 2021-01-12
site: add Milas

milas:
Screen Shot 2021-01-12 at 12 41 38 PM


nicks (approved):
lgtm!

milas:
Just realized I did not touch the locations copy - not sure what makes most sense here as adding another location might start to make things unwieldy.

about page:
> We’re small (9 people, VC-backed). We’re based in NYC, but also have teammates in Berlin, Germany and Richmond, VA.

footer:
> Built with 🌈 from New York, Virginia, Berlin and the ☁️

Thoughts? Can also leave it as-is, not a big deal to me.

nicks:
Haha, we're also missing that one person moved to Philly. Let's leave it alone for now and I can come back and rewrite it later.

PR #798, opened 2021-01-14
docs: add API reference for env for local & local_resource

milas:
See tilt-dev/tilt#4072 for corresponding feature implementation.

### api docs: `local`
docs_api_local

### api docs: `local_resource`
docs_api_local_resource
NOTE: this also adds `tags` to the function arg signature but doesn't add any actual docs around it besides that (fyi @maiamcc)

### example docs: `local_resource` w/ env
docs_example_local_resource_env
NOTE: there's also a slight correction to add a `name` to the other examples in this section, which were missing it and so weren't actually syntactically valid


milas:
This is good to go now that v0.18.5 got cut which includes this 🎉

docs/local_resource.md
```diff
@@ -117,10 +117,20 @@ With `serve_cmd`, when the resource updates:
...
-``local_resource(serve_cmd='kubectl port-forward -n openfaas svc/gateway 8080:8080')``
+``local_resource('kube-port-forward', serve_cmd='kubectl port-forward -n openfaas svc/gateway 8080:8080')``
 
 #### show the k8s api server's logs
-``local_resource(serve_cmd='kubectl logs -f -n kube-system kube-apiserver-docker-desktop')``
+``local_resource('kube-logs', serve_cmd='kubectl logs -f -n kube-system kube-apiserver-docker-desktop')``
+
+#### custom environment
+```python
+update_env={'CGO_ENABLED': '0'}
```
maiamcc, line 130: not necessary, but if you want to be extra clear, consider inline comments:
```
update_env={'CGO_ENABLED': '0'}  # will override any values for this key currently in the env
serve_env={'PORT': os.getenv('PORT', default='8001')}  # if $PORT set, use that value, else use '8001'
```

docs/local_resource.md
```diff
@@ -117,10 +117,20 @@ With `serve_cmd`, when the resource updates:
 Some examples:
 
 #### build and run a server locally
-``local_resource(cmd='go build ./cmd/myserver', serve_cmd='./myserver --port=8001', deps=['cmd/myserver'])``
+``local_resource('local-server', cmd='go build ./cmd/myserver', serve_cmd='./myserver --port=8001', deps=['cmd/myserver'])``
```
maiamcc, line 120: gooood catch 😅

api/api.py
```diff
@@ -881,7 +886,10 @@ def local_resource(name: str, cmd: Union[str, List[str]],
                    auto_init: bool=True, serve_cmd: str = "", cmd_bat: str = "",
                    serve_cmd_bat: str = "",
                    allow_parallel: bool=False,
-                   links: Union[str, Link, List[Union[str, Link]]]=[]) -> None:
+                   links: Union[str, Link, List[Union[str, Link]]]=[],
+                   tags: List[str] = [],
```
maiamcc, line 890: oh whoooops this is supposed to be a secret parameter for now, but I see why you need to mention it in the function signature... it's proooobably fine to mention it here and just not in the list of params below? 🤞

maiamcc (approved):
LGTM! (ty for re-requesting review, this totally fell off my radar.)

docs/local_resource.md
milas: Good call - I put them on their own lines just to prevent horizontal scroll at most browser widths (we get about 65 cols max or so before it'll force a scroll)

Screen Shot 2021-01-25 at 9 42 51 AM

api/api.py
milas: Yeah, I can't imagine anyone is using positional args this deep in a `local_resource` definition so it really shouldn't effect anyone either way...I almost omitted it but erred on the side of staying accurate

PR #801, opened 2021-01-15
docs: update asdf instructions for 0.18.4

jazzdan:
I think this one was missed :D


nicks:
thanks @jazzdan ! We updated the automation script instead

PR #803, opened 2021-01-22
install/upgrade: bump tilt version numbers for asdf in upgrade.md

maiamcc:
script is fixed in https://github.com/tilt-dev/tilt/pull/4110 but we missed it this time around (ty @jazzdan for https://github.com/tilt-dev/tilt.build/pull/801/ !)


nicks:
obsolete?

PR #806, opened 2021-01-29
docs: add CRDs to the list of live_update exceptions

landism:
😢


maiamcc (approved):
Sigh :(

Can you update the extension readme too?

PR #810, opened 2021-02-03
docs: fix default value for disable in docker_prune_settings

milas:
The default here is `False` (i.e. enable Docker prune); the docs
were incorrect.


maiamcc (approved):
😅  good catch!

PR #816, opened 2021-02-10
Revert "tilt-site: Remove contact us page since we don't need it anym…

landism:
#815 deleted this because "we don't need it any more", but it's still linked from many places:

https://github.com/tilt-dev/tilt.build/search?q=tilt.dev%2Fcontact

### DIFF
--- a/src/contact.md
+++ b/src/contact.md
@@ -1,0 +1,84 @@
+---
+layout: page
+title: Contact Us
+permalink: /contact
+---
+
+<div class="u-marginBottomUnit">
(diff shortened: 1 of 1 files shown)


nicks (approved):

victorwuky:
@landism : Thanks for finding these. I'll go to all the places and replace with mail links like https://github.com/tilt-dev/tilt.build/pull/814, before re-deleting this page. https://app.clubhouse.io/windmill/epic/11234

PR #817, opened 2021-02-10
Replace contact us links with mail links

victorwuky:


victorwuky:
Let's at least do this first: https://github.com/tilt-dev/tilt.build/pull/818

PR #820, opened 2021-02-12
docs: secret page for Tests in Tilt [ch11093]

maiamcc:
Hello @ellenkorbes, @landism,

Please review the following commits I made in branch maiamcc/tests-docs:

8563d3d2d7ad3f8ad8731381a2bc9ab3c555f5f6 (2021-02-12 12:25:04 -0500)
add images, tweak copy

73aa40ae6aa633f90c33c8a2f919eea02d8e9bdc (2021-02-11 17:13:17 -0500)
flesh out snippets

edae414e96c42a608048f1bdca473a5e1faf3745 (2021-02-09 17:42:48 -0500)
outline


docs/tests_in_tilt.md
```diff
@@ -0,0 +1,109 @@
...
+# A test depending on specific relevant files
+test('unit-test-remittance', 'pytest remittance_test.py',
+     deps=['app/remittance.py', 'app/remittance_test.py', 'app/payment_utils.py'])
+```
+(More on how to intelligently set your dependencies below.)
+
+Sometimes, you don't want your test(s) to execute automatically. Maybe you want to turn off automatic-ness for all but a few tests you're iterating on, or prevent an expensive integration test from running all the time. Use the "auto" trigger in the UI to toggle automatic execution on or off for a given test.
+
+<figure>
+    <img src="/assets/img/tests-in-tilt/auto-toggles" alt="a close-up of several cards in the sidebar, with 'auto' variously toggled on or off">
```
ellenkorbes, line 51: Missing file extension?

docs/tests_in_tilt.md
```diff
@@ -0,0 +1,109 @@
...
+If your whole unit test suite runs relatively quickly/has caching (like Go tests), you can just run the whole thing when relevant files change:
+```python
+def all_go_files(path):
+  return str(local('find . -type f -name "*.go"')).split("\n")
+
+test('go-tests', 'go test ./... -timeout 30s', deps=all_go_files())
+```
+
+#### Go tests by package
+This approach gives you a little more granularity:
```
ellenkorbes, line 70: ELI5 for the kids at the back maybe? "This way your tests will be split into different cards" or something?

ellenkorbes (approved):
This is perfect Maia, thank you!

docs/tests_in_tilt.md
```diff
@@ -0,0 +1,109 @@
+---
+title: Tests in Tilt
+description: "Run your tests from Tilt for visibility, responsiveness, and fine-grained control."
+layout: docs
+---
+With the Tilt `test` primitive, you can run your tests from Tilt itself. Take advantage of Tilt's file-watching and responsiveness, and the power and customizability of Starlark, to supercharge your testing workflow.
```
hyu, line 6: I'd say for the last bit, something like "...to include tests with the rest of your workflow in Tilt."
Just to be more modest and descriptive. What do you think?

landism: +1 for modesty

docs/tests_in_tilt.md
```diff
@@ -0,0 +1,109 @@
...
+(More on how to intelligently set your dependencies below.)
+
+Sometimes, you don't want your test(s) to execute automatically. Maybe you want to turn off automatic-ness for all but a few tests you're iterating on, or prevent an expensive integration test from running all the time. Use the "auto" trigger in the UI to toggle automatic execution on or off for a given test.
+
+<figure>
+    <img src="/assets/img/tests-in-tilt/auto-toggles" alt="a close-up of several cards in the sidebar, with 'auto' variously toggled on or off">
+    <figcaption>In this example, tests "foo" and "beep" are in auto mode--they will execute automatically when any of their `deps` are changed.</figcaption>
+</figure>
+
+You can also set default AUTO values for your tests via the Tiltfile, using [the `trigger_mode` parameter](https://docs.tilt.dev/manual_update_control.html).
```
landism, line 55:
```suggestion
You can also control this behavior via the Tiltfile, using [the `trigger_mode` parameter](https://docs.tilt.dev/manual_update_control.html).
```

landism: maybe a bit less awkward?

docs/tests_in_tilt.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Programmatically Registering Tests
+Part of what's great about Starlark (the dialect of Python that Tiltfiles are written in) is that it's a _programming language_, so you can use it for complex stuff. In particular, you can use it to programmatically register your tests to Tilt in whichever way works best for you. Here are some examples:
```
landism, line 58:
```suggestion
Part of what's great about Starlark (the dialect of Python that Tiltfiles are written in) is that it's a _programming language_, which makes Tiltfiles extremely flexible. Let's look at a few of the many ways you can register your tests to Tilt, according to your tastes:
```

I don't have strong feelings on this but some vague senses of things that I tried to correct:
1. maybe just me but not a huge fan of "stuff" in this context
2. maybe better to not give the impression that what's coming up is "complex"
3. I think this usage of "in particular" is wrong, since there are lots of great complex things you can do in starlark that are not registering tests. I get the sense of what you're going for and maybe something like "in particular as far as tests are concerned, you can..." could work, but feels awkward. But also I don't know if anyone will really notice or care and I don't even know if I'd comment on this if anyone but you had written it.

docs/tests_in_tilt.md
landism: +1

also hit on the change in test run behavior, like "Each package's tests will only run when that package changes"

NB this also has a key distinction from the previous approach: `go test` does have good caching, but a test's results will be invalidated when one of its dependencies change. If package `foo` depends on `bar`. With the first approach, when `bar` changes, we re-rerun both `foo` and `bar`'s tests, and neither is cached. With the second approach, when `bar` changes, we do not re-run `foo`'s tests, since we didn't add `bar` to its deps. This speeds things up but also potentially misses breaks.

docs/tests_in_tilt.md
maiamcc: 😅

docs/tests_in_tilt.md
```diff
@@ -0,0 +1,109 @@
...
+def pretty_name(s):
+  return s.replace(CWD, '').lstrip('/')
+
+for pkg in all_go_package_dirs():
+  name = pretty_name(pkg)
+  test(name, 'go test %s' % pkg, deps=[pkg])
+```
+
+#### JS tests by test file
+This snippet naively matches by prefix--e.g. `foo_test.tsx` will run on changes to `foo.tsx` and `foo_test.tsx`. Note that it's written for a flat file hierarchy (i.e. all the JS files and tests live at the root of `web/src`, but is easy to modify to fit your directory structure).
```
landism:
```suggestion
This snippet naively matches by prefix--e.g. `foo.test.tsx` will run on changes to `foo.tsx` and `foo.test.tsx`. Note that it's written for a flat file hierarchy (i.e. all the JS files and tests live at the root of `web/src`, but is easy to modify to fit your directory structure).
```

landism (approved):
LGTM

docs/tests_in_tilt.md
maiamcc: i love this comment thank you for this comment i'm glad you left it

docs/tests_in_tilt.md
```diff
@@ -0,0 +1,109 @@
...
+# A test depending on an entire directory
+test('cart-tests', 'go test ./pkg/cart', deps=['./pkg/cart'])
+
+# A test depending on specific relevant files
+test('unit-test-remittance', 'pytest remittance_test.py',
+     deps=['app/remittance.py', 'app/remittance_test.py', 'app/payment_utils.py'])
+```
+(More on how to intelligently set your dependencies below.)
+
+Sometimes, you don't want your test(s) to execute automatically. Maybe you want to turn off automatic-ness for all but a few tests you're iterating on, or prevent an expensive integration test from running all the time. Use the "auto" trigger in the UI to toggle automatic execution on or off for a given test.
```
maiamcc, line 48: > automatic-ness
um HALP what do @landism

landism: "Maybe you want to iterate on a few tests and disable the rest"?

PR #824, opened 2021-02-17
docs: api docs for os.path.relpath

maiamcc:
docs for https://github.com/tilt-dev/tilt/pull/4207

words are hard -- especially having a hard time explaining the args. halp?

### DIFF
--- a/api/modules/os/path.py
+++ b/api/modules/os/path.py
@@ -6,6 +6,17 @@ def abspath(path: str) -> str:
 """
   pass
 
+def relpath(basepath: str, targpath: str) -> str:
+  """Return the path of `targpath` relative to `basepath`. On success, the returned path will
+  always be relative to `basepath`, even if `basepath` and `targpath` share no elements. An
+  error is returned if `targpath` can't be made relative to `basepath`.
+
+  Args:
+    basepath: Path to which ``targpath`` will be made relative
+    targpath: A filesystem path
+"""
+  pass
+
 def basename(path: str) -> str:
   """Return the basename of the path.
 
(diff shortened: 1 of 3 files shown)


api/modules/os/path.py
```diff
@@ -6,6 +6,17 @@ def abspath(path: str) -> str:
 """
   pass
 
+def relpath(basepath: str, targpath: str) -> str:
+  """Return the path of `targpath` relative to `basepath`. On success, the returned path will
+  always be relative to `basepath`, even if `basepath` and `targpath` share no elements. An
+  error is returned if `targpath` can't be made relative to `basepath`.
```
landism, line 12:
```suggestion
  error is raised if `targpath` can't be made relative to `basepath`.
```

landism: Or "it's a tiltfile load error" or something. We aren't very consistent in how we describe this :(.

api/modules/os/path.py
```diff
@@ -6,6 +6,17 @@ def abspath(path: str) -> str:
...
+
+  Args:
+    basepath: Path to which ``targpath`` will be made relative
+    targpath: A filesystem path
```
landism, line 16: Maybe "The path to be made relative to ``basepath``"? "The path to be relativized"? FWIW, https://www.geeksforgeeks.org/python-os-path-relpath-method/ pretty much just lists them both as "a path" and then lets the rest of the docs explain their meaning, rather than trying to explain each in isolation

and/or just say "whatever these words are going to be awkward regardless" and add some examples worth a thousand words, which I think would end up being much clearer and faster to read.

landism (approved):
I feel like swapping the args would be better though - https://github.com/tilt-dev/tilt/pull/4207#discussion_r577994761

landism (approved):

PR #828, opened 2021-03-04
blog: a post on tilt-apiserver

nicks:
Hello @ellenkorbes, @milas,

Please review the following commits I made in branch nicks/curl:

550152aedb669557b2251cb4011e319422cf38e1 (2021-03-04 12:55:59 -0500)
blog: a post on tilt-apiserver


nicks:
Mainly trying to externalize a lot of the discussions we've been having internally about extensibility and why it's useful

milas (approved):

ellenkorbes (changes requested):
As promised, nits:

1. Who's the audience? I don't know who we're writing this for. Is it for ourselves? For our biggest fans who want a peek behind the curtain? For kube-nerds after hours at the bar? You don't have to answer *me,* but the fact that I can't tell is a concern.

2. There's a knowledge level mismatch throughout parts of the article. I.e. 
  - I wouldn't expect that someone who's capable of even installing Kind needs to be told what a pod is.
  - When I see the explanation of what a pod is I wanna stop reading. It makes me assume there's nothing for me to learn here. But soon after there's some interesting, non-obvious stuff like `kubectl -v 6` that I didn't know and am glad to have learned.

3. `We think building servers this way could benefit a lot of tools.` This should be at the beginning, rephrased. You don't really tell me at the top what I'll be reading about in concrete terms. I feel like I was trying to figure out what the point was the whole time, until I read this and it all fell into place. So I'd suggest moving it up.

It's an interesting article otherwise. Interesting and useful bits; but the current framing gets in the way.

nicks:
ha - i think the line about pod is just badly phrased - not trying to explain what a pod is. Just trying to establish the analogy between the pod API and /proc. I'll rephrase.

re: the audience - trying to capture the tone / technical level of https://wizardzines.com/comics/containers-arent-magic/, which i really admire. i.e., developers who use Kubernetes, but treat it as a black box, and would benefit from a deeper understanding of how it works.

re: the conclusion line - ya, i agree with that, will move it up

PR #840, opened 2021-04-06
docs: add docs on how to set up tilt on ci

nicks:
Hello @maiamcc, @ellenkorbes,

Please review the following commits I made in branch nicks/ci3:

e3cfdb6d20fc8efafd1792f43a5dd5fe7ed67bb6 (2021-04-06 11:13:20 -0400)
docs: add docs on how to set up tilt on ci
This basically cannibalizes the blog post
https://blog.tilt.dev/2021/04/02/kubernetes-on-ci.html
and other CI docs into a central docs, in the spirit
of cannibalizing content more often

### DIFF
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -1,0 +1,152 @@
+---
+title: With CI
+description: "Test that your images build and your servers start"
+layout: docs
+---
+
+Once you've bought into container-based development,
+you'll want to set up CI to ensure your servers don't break.
+
+We think of this in two pieces:
+
+1) Run Tilt in CI to deploy to a single-use cluster.
+
+2) Run a single-use Kubernetes in CI that you can immediately throw away.
+
+Let's dig into examples of each one.
+
+## The `tilt ci` command
+
+The [`tilt ci` command](/cli/tilt_ci.html) is made for CI jobs.
+
+### How it Works
+
+Under the hood, `tilt ci`:
+
+1) Executes your Tiltfile.
+
+2) Runs all `local_resource` commands.
+
+3) Builds all images.
+
+4) Deploys all Kubernetes resources.
+
+5) Waits until all servers and other Kubernetes resources are healthy.
+
+`tilt ci` defaults to log-streaming mode. The web UI is still accessible,
+but the terminal will be a simple log stream.
+
+If any step fails, Tilt will exit immediately with an error code.
+
+If any resource crashes or looks like it will never succeed (e.g., a pod that
+takes too long to schedule), Tilt will also exit with an error code.
+
+Once all services are healthy, it will exit with status code 0.
+
+### Examples
+
+Most of our example projects use CircleCI to run `tilt ci`:
+
+<ul>
+  {% for page in site.data.examples %}
+     <li><a href="/{{page.href | escape}}#ci">{{page.title | escape}}</a></li>
+  {% endfor %}
+</ul>
+
+These each use [`ctlptl`](https://ctlptl.dev/) to set up a single-use
+cluster. But if you're considering other options, read on.
+
+## Single-use Kubernetes clusters
+
+If you're using Kubernetes for dev, you'll likely also want to use Kubernetes in CI.
+
+Tilt-team has this problem as well! Our integration test suite uses the latest
+version of Tilt to deploy real sample projects against a real cluster.  So we
+have a lot of experiences with best practices.
+
+Here are some options, with the pros and cons of each:
+
+### Easiest: Cluster on VM-based CI
+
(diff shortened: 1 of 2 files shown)


docs/ci.md
```diff
@@ -0,0 +1,152 @@
...
+</ul>
+
+These each use [`ctlptl`](https://ctlptl.dev/) to set up a single-use
+cluster. But if you're considering other options, read on.
+
+## Single-use Kubernetes clusters
+
+If you're using Kubernetes for dev, you'll likely also want to use Kubernetes in CI.
+
+Tilt-team has this problem as well! Our integration test suite uses the latest
```
maiamcc, line 63:
```suggestion
Team Tilt has this problem as well! Our integration test suite uses the latest
```
or "The Tilt Team"? Idk this just seems weird

docs/ci.md
```diff
@@ -0,0 +1,152 @@
...
+
+Tilt-team maintains [`ctlptl`](https://ctlptl.dev/), a CLI for declaratively
+setting up local Kubernetes clusters. If you're using `ctlptl`, it will try to
+detect when you have a remote docker environment and set up the `socat`
+forwarding automatically.
+
+If you want to wire it up yourself, check out
+[this Bash script](https://github.com/tilt-dev/kind-local/blob/master/.circleci/with-kind-cluster.sh).
+
+### Not Recommended: Remote Registry
```
maiamcc, line 125: might put this section earlier, so the "kind|minikube lets you use a local registry!" points have more context/are more impressive

docs/ci.md
```diff
@@ -0,0 +1,152 @@
...
+
+Tilt-team has this problem as well! Our integration test suite uses the latest
+version of Tilt to deploy real sample projects against a real cluster.  So we
+have a lot of experiences with best practices.
+
+Here are some options, with the pros and cons of each:
+
+### Easiest: Cluster on VM-based CI
+
+[`kind`](https://kind.sigs.k8s.io/) is currently the gold standard for running
```
maiamcc, line 71: nbd but I like adding a little context -- consider some/all of the below
```suggestion
[`kind`](https://kind.sigs.k8s.io/) (Kubernetes in Docker, which runs a single-node Kubernetes cluster as a Docker container) is currently the gold standard for running
```

docs/ci.md
```diff
@@ -0,0 +1,152 @@
...
+
+Most of our example projects use CircleCI to run `tilt ci`:
+
+<ul>
+  {% for page in site.data.examples %}
+     <li><a href="/{{page.href | escape}}#ci">{{page.title | escape}}</a></li>
+  {% endfor %}
+</ul>
+
+These each use [`ctlptl`](https://ctlptl.dev/) to set up a single-use
```
maiamcc, line 56: consider a brief description of ctlptl (we do that lower down but not here?)

docs/ci.md
```diff
@@ -0,0 +1,152 @@
...
+---
+
+Once you've bought into container-based development,
+you'll want to set up CI to ensure your servers don't break.
+
+We think of this in two pieces:
+
+1) Run Tilt in CI to deploy to a single-use cluster.
+
+2) Run a single-use Kubernetes in CI that you can immediately throw away.
```
maiamcc, line 14: Hm, I'm not sure this gets across the relationship between the steps. Consider something like:
```
`tilt ci` is our recommended way to make sure your servers haven't broken from your latest change--it spins up your whole Tilt config and makes sure everything is healthy. To run this as part of your CI pipeline, you'll need a Kubernetes cluster to run it against--ideally a single-use cluster that can be easily thrown away when you're done.

Thus, we have two things to dig into here:
1. how to use `tilt ci` to make sure your servers are healthy
2. how to get access to a Kubernetes cluster from CI that you can run `tilt ci` against
```

maiamcc (approved):

docs/ci.md
nicks: hmmm...what context did you have in mind? it seems weird to say "these use ctlptl (a tool for setting up clusters) to set up a single-use cluster" (i.e., the description of what it does seems to already be here)

docs/ci.md
nicks: ya, i think this is a good criticism - rewrote it a bit

docs/ci.md
nicks: added some detail. also it's no longer true that kind clusters are single-node!
https://pkg.go.dev/sigs.k8s.io/kind/pkg/apis/config/v1alpha4#Cluster

nicks (commented):
thanks!

docs/ci.md
maiamcc: 🤯

PR #852, opened 2021-05-05
Update features.yml

PeterChapman:
"Very" does nothing for us here other than to make this sound clumsy. "Holistic" alone is cleaner and more efficient.


hyu (approved):
✨

landism:
this always bugged me!

PR #854, opened 2021-05-12
site: replace enterprise page with a 'tilt shift' page

nicks:
Hello @hyu,

Please review the following commits I made in branch nicks/cta:

d2aae90e2af2b093f8a473520f045aec92338266 (2021-05-12 18:22:10 -0400)
site: replace enterprise page with a 'tilt shift' page

### DIFF
--- a/src/_data/footer.yml
+++ b/src/_data/footer.yml
@@ -1,5 +1,5 @@
-- name: Enterprise
-  href: enterprise
+- name: Schedule a Guide
+  href: liftoff
   primary: true
   landingurl: true
 - name: Docs
--- a/src/_data/github.yml
+++ b/src/_data/github.yml
@@ -1,2 +1,2 @@
-stars: 4.2k
+stars: 4.3k
 href: https://github.com/tilt-dev/tilt
--- a/src/_data/header.yml
+++ b/src/_data/header.yml
@@ -1,5 +1,5 @@
-- name: Enterprise
-  href: enterprise
+- name: Schedule a Guide
+  href: liftoff
   primary: true
   landingurl: true
 - name: Docs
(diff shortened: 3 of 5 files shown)


nicks:
After a bunch of discussions with @PeterChapman and L, we're going to change our enterprise engagements from more open-ended support to something much more time-bound and focused on getting the initial Tiltfile setup.

I don't want to get too precious about this, so I took the existing enterprise page and adapted Peter's marketing copy here: https://docs.google.com/document/d/1V_hVzC96Bf_CgZI_-4YOolv_ck5WKOYgwt8xxE4f3NI/edit

I also changed it from "Tilt Liftoff" to "Tilt Shift" which is IMO a way better pun
https://www.netapp.com/knowledge-center/what-is-lift-and-shift/
https://en.wikipedia.org/wiki/Tilt%E2%80%93shift_photography

PeterChapman:
Thanks for doing this! I'm not going to cancel the wedding over this, but I strongly prefer 'liftoff' over 'shift'. Liftoff, to me, has strong connotations that communicate what we're offering customers: it's fast, temporary, and it's the beginning of something. It's literally what you do to launch something. 

'Tilt shift' is clever, but it doesn't communicate what is behind the offering.

hyu (approved):
Makes sense! This is also a good reminder to me that this page exists. (I had a recent customer discussion where I could have used this link...)

nicks:
@PeterChapman ya, that's fair. Liftoff, to me, has an early 2010s vibe that feels a bit overused/dated, when we're really trying to shift to being more surprising/winking. But will change it back to liftoff and see if I can come up with something better. 

Was also playing with "Tilt Around and Find Out" or https://en.wikipedia.org/wiki/Tiltrotor

nicks:
oh wait how did i forget Tilt-a-Whirl!!! https://en.wikipedia.org/wiki/Tilt-A-Whirl
got to find a way to use that

PR #855, opened 2021-05-13
entr: a blog post

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/entr:

7ee728dcb12f79e35bfc7b09d302ec4dab319d2f (2021-05-12 20:28:22 -0400)
entr: a blog post


nicks:
this is a hack week demo of how to build 'entr' on tilt, but more in a narrative. will get it reviewed next week if we want to publish it.

nicks:
@ellenkorbes wdyt?

ellenkorbes (approved):
LGTM

ellenkorbes:
As feedback for next time, we could expand on the story, on who would want to implement their own entr, and why, and how, and show an example of that real-world scenario.

But that would possibly defeat the purpose of these posts being short and quick to create, so take this feedback with a bucketful of salt.

nicks:
ya, i think that's right...I think a good tone here is "we're going to mix new tech with very old tech", like in your memory card video

PR #860, opened 2021-05-21
(chore) Add Lizz to About page and update Han's picture

lizzthabet:
@hyu, let me know if this photo crop is ok!

Screen Shot 2021-05-21 at 11 37 27 AM


ellenkorbes (approved):
this reminds me mine is super out of date 😬

.gitignore
```diff
@@ -10,6 +10,7 @@
 /test/tmp/
 /test/version_tmp/
 /tmp/
+.DS_Store
```
hyu, line 13: Should this be part of the commit? I also hate these invisible Mac files but I think I set my system global gitignore to exclude these, since they're less specific to this project.

hyu (approved):
Thanks for helping me update my photo, at long last!

.gitignore
lizzthabet: Gotcha, I can do that!

PR #868, opened 2021-06-09
docs: remove auto_init=False / TRIGGER_MODE_AUTO restrictions

milas:
Previously, you could not have a resource that had `auto_init=False`
with `trigger_mode=TRIGGER_MODE_AUTO`. This restriction was lifted
in v0.19.1 a while back by https://github.com/tilt-dev/tilt/pull/4292.

This removes mention of the restriction from the docs including
retconning a blog post from 2019.

### DIFF
--- a/docs/local_resource.md
+++ b/docs/local_resource.md
@@ -73,19 +73,17 @@ but do not write.
 
 ## auto_init
 
-By default, a local resource will run on startup. To disable this behavior, put the
-resource in `TRIGGER_MODE_MANUAL` and specify `auto_init=False`:
+By default, a local resource will run on startup. To disable this behavior, specify `auto_init=False`:
 ```python
-local_resource('reset-db', cmd='reset_db.sh',
-    trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False
-)
+local_resource('reset-db', cmd='reset_db.sh', auto_init=False)
 ```
 
-For more on trigger mode, [see the docs](https://docs.tilt.dev/manual_update_control.html).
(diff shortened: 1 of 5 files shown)


docs/local_resource.md
```diff
@@ -73,19 +73,17 @@ but do not write.
...
-    trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False
-)
+local_resource('reset-db', cmd='reset_db.sh', auto_init=False)
 ```
 
-For more on trigger mode, [see the docs](https://docs.tilt.dev/manual_update_control.html).
+You can combine this with `trigger_mode=TRIGGER_MODE_MANUAL` so the resource only runs when
+explicitly triggered or `trigger_mode=TRIGGER_MODE_AUTO` so that the resource does not run
+at first start but will still run if any of its file dependencies are changed (this can be
+useful for tests, for example).
```
maiamcc, line 84: copy suggestion, feel free to ignore:
```
You can combine this with `trigger_mode=TRIGGER_MODE_MANUAL` (so the resource only runs when
explicitly triggered) or `trigger_mode=TRIGGER_MODE_AUTO` (so that the resource does not run
on "tilt up" but will run if any of its file dependencies are changed--this can be
useful for tests, for example).
```
(bonus points, link "tests" to the test docs -- if they're live? if they're not live, then maybe don't mention tests in this doc.)

maiamcc (approved):
ah, much more thorough than mine, ty!

docs/local_resource.md
milas: Revised as part of cfa70c6

PR #872, opened 2021-06-14
add command working directory for local/local_resource

lucieleblanc:
Accompanying [tilt/#4644](https://github.com/tilt-dev/tilt/pull/4644/files), following [tilt/#4549](https://github.com/tilt-dev/tilt/issues/4549).


milas (approved):
Changes look good, but you will need to either run `tilt up` or `make api` (same outcome - Tilt will just auto run this) and commit the generated HTML as well (`src/_includes/api/functions.html`) -- that doesn't get auto-generated as part of CI right now 😬 

Also, we should wait to merge until the release is tagged (we usually release every Friday) so that it's not live before the code.

PR #885, opened 2021-07-06
blog: add execution clarification for uibutton

milas:
Was asked about this on Twitter - blog post is vague here! (The
extension docs also didn't do a great job of clarifying this and
are being updated too.)

This just adds a note that the execution happens locally (not in
a Pod/container) and an extra call-out to the extension docs.


ellenkorbes (approved):
LGTM!

PR #887, opened 2021-07-08
blog: UIButton pt2: attack of the nav

milas:
Second blog post on UIButton and the extension that goes into
navbar buttons and icons.

It also includes a bit of a primer on the underlying API primitives
at the end in a "Deep Dive" section.


milas:
Netlify preview: https://deploy-preview-887--cranky-newton-c5897d.netlify.app/2021/07/08/uibutton-navbar.html

@ellenkorbes Let me know if you think the "Deep Dive" is awkward - I thought since the navbar buttons are _very_ similar to the resource buttons, it made sense to use this as a way to gently introduce how parts of the API fit together, but provided a natural "fall off" point for anyone who just wanted the how-to and doesn't care about internals.

FYI I also adapted Nick's hackathon script to make a `mailto` instead of pre-fill a Google Doc form because that process is unfortunately not super intuitive (you have to generate a pre-filled version and then find the random parameter IDs to match up and put in Python code) and I didn't want it to become a "how to Google Docs" blog post 😛

ellenkorbes (approved):
Love your writing, this was a lot of fun to read. Ship it!

PR #888, opened 2021-07-08
docs: add subnav with top reference pages

milas:
We've got a LOT of docs, but most of the time when folks end up
on the docs site, they probably want the API or CLI reference,
which are waaaay at the bottom.

This adds a subnav that includes a few links to the (hopefully)
most important pages. The search comes along for a ride out of
the sidebar as well.

There were some...interesting things going on with the layout,
so there's some general changes here to flexbox more flexbox-y,
but they should not actually change things, just make it more
friendly for future adaptions (like this one).

I did a sanity pass across all the pages on the docs site as well
as a few blog posts and whatever I could find on the "main" site.

Thanks to @lucieleblanc for proposing this! 🙌

### DIFF
--- a/src/_layouts/docs.html
+++ b/src/_layouts/docs.html
@@ -3,10 +3,10 @@ layout: default
 ---
 {% assign page_href = page.url | slice: 1, page.url.size %}
 <div class="wrapper wrapper--docs u-marginBottom1_5">
-  <nav class="subnav">
+  <nav class="subnav" aria-label="Most Popular">
     <form action="/search" method="GET">
       <div class="formItem">
-        <input type="text" name="q" id="q" placeholder="Search" />
+        <input type="text" name="q" id="q" placeholder="Search" aria-label="Search Documentation" />
       </div>
     </form>
     <ul class="subnav-links">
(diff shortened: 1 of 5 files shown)


src/_layouts/docs.html
```diff
@@ -1,16 +1,22 @@
 ---
 layout: default
 ---
+{% assign page_href = page.url | slice: 1, page.url.size %}
 <div class="wrapper wrapper--docs u-marginBottom1_5">
-  <nav class="sidebar">
-    <h2><a href="/">Docs</a></h2>
-
+  <nav class="subnav">
```
lizzthabet, line 6: could you add a description in an `aria-label` property to both of the navs? i think when you have multiple `nav`s on a page, it's best practice to have a title for each.

lizzthabet (approved):
lgtm! the new subnav links don't _scream_ navigation to me, but i think it's definitely an improvement and helpful for orienting someone on the high-level important sections of tilt docs.

src/_layouts/docs.html
milas: Ahh thanks for catching that - I named them "Most popular" and "Primary" in 2618e98

lizzthabet: yeah! i think that's totally fine. i tend to be a little more descriptive in my aria descriptions, like "Primary nav for Tilt docs," but i don't have a sense if that level of detail is significant for screen-reader users because there's a lot of other page context that can provide that information.

PR #897, opened 2021-07-30
Fix Kind Setup Link

gaffneyd4:
Previous value https://github.com/tilt-dev/ctlptl#kind-with-a-built-in-registry is not a valid anchor on the README anymore.


nicks (approved):
thanks!

PR #900, opened 2021-08-04
blog: add post for resource grouping

lizzthabet:
i could probably edit this forever, but i'll stop now 😅 

@ellenkorbes, i pulled in a couple of examples from your first round of feedback. let me know if those work or not!


ellenkorbes:
@lizzthabet Only nit on the text is that...

> so that all frontend, backend, and infrastructure services are grouped together

...makes it sound like all three labels will be grouped together in the same bucket. Other than that, 10/10 on the text.

The GIFs though, I feel they're a bit too long. I feel we should re-cut them and make them either embedded YouTube videos or shorter GIFs. Do you have a feeling for which one would be better? Either way I can do the editing, do you have the original video files?

lizzthabet:
> @lizzthabet Only nit on the text is that...
> 
> > so that all frontend, backend, and infrastructure services are grouped together
> 
> ...makes it sound like all three labels will be grouped together in the same bucket. Other than that, 10/10 on the text.

ah, good point! i will rephrase that.

PR #906, opened 2021-08-16
docs: expandable nav sections in sidebar

milas:
This uses a small JS library: https://github.com/nico3333fr/van11y-accessible-accordion-aria/
It follows the general recommended approach online for keeping it
accessible and isn't super opinionated re: styling.

If the user has Javascript disabled, the site will look and function
exactly as it does today. Hurray!

If the user has Javascript enabled, the sidebar nav sections will
become collapsible. All sections other than the active section for
the current page will be collapsed on load.

collapsible nav demo


src/_sass/layout.scss
```diff
@@ -569,6 +569,34 @@ h3.relatedHeader {
...
+.nav-accordion__panel[aria-hidden=true] {
+  display: none;
+}
+
+.nav-accordion__header[aria-expanded="true"]:before {
+  content: "- ";
+}
+
+.nav-accordion__header[aria-expanded="false"]:before {
+  content: "+ ";
```
lizzthabet, line 585: (optional nitpick) there's a little bit of layout shift that occurs with the character change from `+` to `-` and vice versa. i was able to avoid this by adding `font-family: monospace;` to both `:before` classes. removing the additional space after each character will maintain the same visual spacing you have now.

lizzthabet (approved):
this looks great! the accordion library you chose does a great job of adding the right markup.

milas:
@lizzthabet I made the style change you recommended and merged in latest changes (i.e. sticky!) - do you mind taking a quick peek at https://deploy-preview-906--eager-austin-142694.netlify.app/ and seeing if the sticky + collapsible together look ok to you?

lizzthabet:
@milas, lgtm! it may seem a little funny to users that the expandable state isn't persistent across pages, but that's totally acceptable for how we have the site set up now.

milas:
On persistence: I actually checked out a few random similar style navs on (tech doc) sites and they are generally non-persistent. Not saying that makes it right, but at least it doesn't seem too abnormal!

I DO really like the middle ground that a lot of progressively-enhanced JS static sites seem to have - it's persistent for a single "session" (since no reload occurs if you have JS) but if you hard refresh (or open in new tab) it'll go back to default, i.e. it's not actually saving state in a cookie/local storage.

But for now... 🙈

PR #907, opened 2021-08-16
docs: sticky sidebar nav

milas:
On scroll, the sidebar nav will stay sticky at the top of the
screen.

You can still scroll to the bottom of the content area (i.e. the
footer links) and it _will_ scroll out at that point. This behavior
is pretty consistent with other sticky navs on a few docs sites I
played with.

sticky nav demo

### DIFF
--- a/src/_sass/layout.scss
+++ b/src/_sass/layout.scss
@@ -114,6 +114,12 @@ li {
   flex: 0 0 $sidebar-width;
   margin-right: $spacing-unit;
 
+  // sticky on scroll
+  position: sticky;
+  top: 0;
+  overflow-y: auto;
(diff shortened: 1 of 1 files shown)


lizzthabet (approved):
i have to admit, i did not know about `position: sticky;`. 😱 i'm curious how the stickiness plays with the expandable menu section, but i'm sure i can test that out later!

milas:
They interact nicely IMO! No scrollbar will be visible unless you expand a bunch of sections, but it otherwise behaves the same. You might have noticed the 2 PRs appeared very close together in time - I had the same concern so did the work on them together and then split into separate PRs since they also worked okay separately.

PR #910, opened 2021-08-16
docs: revamped article footer area

milas:
* "Was this helpful?" content box centered
* New "Back to top" link on left
* Moved "Edit on GitHub" link on right

### desktop view
article footer area on desktop

### mobile view
Screen Shot 2021-08-16 at 5 39 40 PM
On mobile, the "Back to top" link gets centered and the "Edit on
GitHub" link vanishes.


src/_layouts/docs.html
```diff
@@ -51,22 +51,26 @@ <h3 class="sidebar-sectionTitle">{{ section.title | escape }}</h3>
...
+    <div class="docsFooter">
     {% include helpful_form.html %}
+    <a id="scrollToTop" href="#">
+      <span aria-hidden="true">↑</span>
+      Back to top
+    </a>
+    {% if page.hideEditButton != true %}
+      <a class="editOnGitHub u-hideOnMobile"
+         href="https://github.com/tilt-dev/tilt.build/tree/master/docs/{{page.name}}">
+        {% svg assets/svg/github.svg width=16 height=16 %}
```
lizzthabet, line 67: (tiny nitpick) maybe worth adding the same semantic markup to this svg asset too? or you could add an `alt="github logo"`.

src/_layouts/docs.html
```diff
@@ -51,22 +51,26 @@ <h3 class="sidebar-sectionTitle">{{ section.title | escape }}</h3>
...
-    {% endif %}
-    
     <h1>{{ page.title | escape }}</h1>
     
     {{ content }}
 
+    <div class="docsFooter">
     {% include helpful_form.html %}
+    <a id="scrollToTop" href="#">
+      <span aria-hidden="true">↑</span>
```
lizzthabet, line 61: (tiny nitpick) because this is an ascii character and accessible to assistive tech, i think you could leave it without any additional markup, but totally up to you! 

not applicable here, but as a sidenote, i like to use `role="presentation"` for icons and elements like this. the difference bw it and `aria-hidden` is that it says to assistive tech "this element is here, but it doesn't have semantic value to you."

src/_layouts/docs.html
```diff
@@ -51,22 +51,26 @@ <h3 class="sidebar-sectionTitle">{{ section.title | escape }}</h3>
...
-    </a>
-    {% endif %}
-    
     <h1>{{ page.title | escape }}</h1>
     
     {{ content }}
 
+    <div class="docsFooter">
     {% include helpful_form.html %}
+    <a id="scrollToTop" href="#">
```
lizzthabet, line 60: i noticed that the footer links are _slightly_ misaligned. if you add `display: block; float: left;` to this link, it should align the items!

Screen Shot 2021-08-17 at 11 19 11 AM

lizzthabet (changes requested):
lgtm! two tiny optional a11y notes and one style change request.

src/_layouts/docs.html
milas: WHY
NOT
BOTH

(6c98e1f)

src/_layouts/docs.html
milas: 😓 (fixed in 6c98e1f)

src/_layouts/docs.html
lizzthabet: hahaha well i wonder if they kind of cancel each other out. like the alt text is there for assistive tech to read, but then you're saying that that element doesn't have any meaning. (or at least, that's what i thought/think. the formal docs on [presentation](https://www.w3.org/TR/using-aria/#presentation) are more complicated than i realized.) tldr; alt text should be good enough!

lizzthabet (approved):
sorry i was not clear on the `role` or `alt` suggestion. 😅 i think choosing one of the other is the way to go!

hyu:
This LGTM. I was musing on [Slack](https://tilt-dev.slack.com/archives/CD9C65ME0/p1629218509002100?thread_ts=1629149991.001100&cid=CD9C65ME0) about a different layout for this section, but I think this PR is already a significant improvement over the existing layout and should proceed post haste!

PR #912, opened 2021-08-18
docs: copy to clipboard + code styling tweaks

milas:
Big addition here is a "Copy" button next to each code block to
copy to clipboard (uses `clipboard.js` - works in all modern
browsers and does not use Flash).

As part of this, the `<code>`/`<pre>` styling was tweaked slightly to improve
spacing: the code blocks are closer to the element that precedes
them, which is logically how the docs flow/intend them to be used.
(For this reason, the button is on the bottom!)

The "Install" page also had its Markdown tweaked to fix the way
the list elements were organized to prevent breaking a list into
multiple, which is both semantically better and helps the spacing
on the content substantially.

### Comparison
BEFORE: left
AFTER: right

#### Install page
install page before & after

#### Tutorial Page
tutorial page before & after

### Demo
copy to clipboard demo


src/assets/js/copy-code.js
```diff
@@ -0,0 +1,52 @@
...
+        const clipboard = new ClipboardJS('button.copy-to-clipboard');
+        clipboard.on('success', (e) => {
+            e.trigger.innerHTML = clipboardSVG + 'Copied!';
+            e.trigger.removeAttribute('aria-label');
+            e.clearSelection();
+
+            // restore the button state after a few seconds
+            setTimeout((target) => {
+                target.innerHTML = copyHTML;
+                target.setAttribute('aria-label', copyLabel);
```
lizzthabet, line 44: i'm curious about toggling the aria markup! is there a reason for that?

lizzthabet (approved):
this looks great! i do wonder if there's a semantic way to link the button to each code block it relates to, but i'm not sure if an `aria-describes` is the best way to do that. no aria is better than bad aria, as the docs say!

src/assets/js/copy-code.js
milas: I thought "Copy to clipboard" was more descriptive for AT (especially since, as you pointed out, there doesn't seem to be a great way to tie this to a specific element) than just "Copy" which is the visible text.

But once you have triggered it, the contents change (temporarily) to "Copied!" which should be sufficient context-wise on its own at that point, but if I left the `aria-label` as-is, it'd be mismatched, so I wanted to make sure the feedback wasn't obscured by the label.

PR #913, opened 2021-08-20
docs: fix search

milas:
Remove the `jekyll-target-blank` plugin which was recently added
in #902 to open all external links in a new window.

Unfortunately, nokogiri (the underlying Ruby lib that parses the
HTML) doesn't support XML namespaced elements properly, so was
converting `<gcse:searchresults-only>` -> `<searchresults-only>`
when processing the DOM to modify outbound links.


lizzthabet (approved):
thanks!!

PR #914, opened 2021-08-20
docs: search styling improvements

milas:
* Populate search box with `q` query param
* Hide "Was this doc helpful?" on search results
* Hide search element if JS disabled
* Show notice that JS is required on search page if disabled
* (Bonus) Hide "Was this doc helpful?" on all pages if JS disabled

docs search page after changes
(I cheated and deleted some results to make this screenshot not 10,000px tall)

### DIFF
--- a/src/_sass/layout.scss
+++ b/src/_sass/layout.scss
@@ -679,6 +679,7 @@ button.nav-accordion__header {
   flex-wrap: wrap;
   align-items: baseline;
   justify-content: space-between;
+  margin-bottom: 1.25 * $spacing-unit;
 
   form {
     flex: 1 0 auto;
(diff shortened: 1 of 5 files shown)


src/_sass/layout.scss
```diff
@@ -1417,7 +1418,6 @@ img.no-shadow {
 .formItem {
   display: block;
   position: relative;
-  padding-bottom: 1.25 * $spacing-unit;
```
milas, line 1420: This moved to be `.subnav` margin as it turns out it's what was providing the spacing and was no longer visible w/o JS so would squish things.

Despite its name `.formItem` is only used for the search element so this won't affect anything else 😅

lizzthabet (approved):
this looks great! the rebase fixed the search functionality for me. do you think it's worth removing the `edit on github` link as well on this page? (not necessary for this pr, but thought i'd mention it since we're already doing conditional rendering for the `was this helpful?` box.)

PR #916, opened 2021-08-23
docs: re-enable links opening in external window

milas:
Re-add the `jekyll-target-blank` plugin but adjust the Google Custom
Search so that it doesn't break it.

The search fix is to use a `<div>` instead of the custom
`<gcse:searchresults-only>` element (which was erroneously having its
`gcse` "namespace" chopped off by nokogiri during DOM processing). This
is actually the way that Google has it in their docs now as of Aug 2021
and works fine.

While it's not great that nokogiri can't handle this properly, the
likelihood we use custom elements in the future with prefixed namespaces
is minimal, and the alternatives are:
 * Use JS to add `target="_blank"` to all links on page load
 * Remember to specify it manually on all external links

### DIFF
--- a/docs/search.html
+++ b/docs/search.html
@@ -15,4 +15,4 @@ layout: docs
      s.parentNode.insertBefore(gcse, s);
  })();
 </script>
-<gcse:searchresults-only></gcse:searchresults-only>
+<div class="gcse-searchresults-only" />
--- a/src/Gemfile
+++ b/src/Gemfile
@@ -23,6 +23,7 @@ group :jekyll_plugins do
   gem "jekyll-paginate"
   gem "jekyll-sitemap"
   gem "jekyll-seo-tag"
+  gem "jekyll-target-blank"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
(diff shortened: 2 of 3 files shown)


milas:
The fix to make these play nice together was actually straightforward 🤦 

Doc reference: https://developers.google.com/custom-search/docs/tutorial/implementingsearchbox

They now use
```html
<div class="gcse-searchresults-only" />
```
instead of
```html
<gcse:searchresults-only></gcse:searchresults-only>
```

I'm guessing they switched to a `<div>` with a class due to compatibility issues like we were hitting since `:` in an tag name is kinda-sorta valid HTML but nobody can really agree it seems (it's NOT a real namespace like in XML I learned!).

lizzthabet (approved):
this is great! haha i will admit to feeling silly for not suggesting this, because i definitely spent a good chunk of friday looking at the google implementation docs with this class name pattern.

the two (non-blocking) things i'll note:
* tilt links from the search results do open in a new window. i don't know if this is intentional because the original pr says that external links should open in a new window.
* there is a hover transition style difference between words that are bolded and not bolded in the search results that i noticed:


(this may have been an existing issue predating some of the recent docs site changes and isn't a huge deal at all.)

milas:
Ahh nice finds! Neither of those regressed/changed with this, so I will open a couple teeny follow-up PRs for them!

PR #918, opened 2021-08-23
site: remove office hours experiment

milas:
We've decided to put this on hold for now!

I reverted to the copy before the introduction, which has the Calendly for user research. If that's not what we want as a substitute for now, just lemme know what makes sense to replace it with!

### DIFF
--- a/src/_includes/index_cta.html
+++ b/src/_includes/index_cta.html
@@ -22,15 +22,6 @@
           <li>Pairing sessions to help you draft and evolve your Tiltfile.</li>
           <li>Help from our content strategists on how to talk about and teach Tilt.</li>
         </ul>
-        <hr/>
-        <p>
-          Not sure if a formal program is right for you?
-        </p>
-        <p>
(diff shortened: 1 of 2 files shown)


src/_includes/index_cta.html
```diff
@@ -22,15 +22,6 @@ <h2 class="modal-title">
           <li>Pairing sessions to help you draft and evolve your Tiltfile.</li>
           <li>Help from our content strategists on how to talk about and teach Tilt.</li>
         </ul>
-        <hr/>
-        <p>
-          Not sure if a formal program is right for you?
```
nicks, line 27: maybe keep this part, but direct them to the Slack channel?

nicks (approved):

src/_includes/index_cta.html
milas: Done!

Screen Shot 2021-08-24 at 5 13 57 PM

PR #928, opened 2021-08-31
blog: Dockerfile optimization tricks

milas:
This was originally from content I'd planned for office hours
that I formalized into a blog post.

It does assume the reader is fairly comfortable with Docker
(+ `Dockerfile`) concepts, but I tried to briefly provide
some context, particularly for `cache` mounts, which are
still pretty unknown despite being usable in stable Docker
w/o weird experimental syntax directives for a while now.


blog/_posts/2021-09-01-docker-does-not-mean-slow.md
```diff
@@ -0,0 +1,146 @@
...
+
+This blog post isn't comprehensive, but focuses on a couple of oft-neglected and lesser-known techniques.
+
+First, we'll cover some common pitfalls that `.dockerignore` can help you avoid.
+Then we'll look at `cache` mounts that let you re-use files between builds even after layers have been invalidated.
+
+Now's a good time to grab a coffee: we're going to dive in head-first!
+
+### Create & Tune Your `.dockerignore`
+[`.dockerignore`][dockerignore] is analogous to `.gitignore`, but instead of excluding files to commit, it excludes files from the image build context.
```
ellenkorbes, line 25: I feel like if we're assuming people are familiar with .gitignore, then they would also be familiar with .dockerignore; and if they're not familiar with .dockerignore they're probably not familiar with .gitignore either? 🤔

ellenkorbes:
OK, the dockerignore/gitignore thing was the only nitpick; this is great, love your writing! :))

ellenkorbes (approved):

blog/_posts/2021-09-01-docker-does-not-mean-slow.md
milas: That's a fair point! I revised this and lost the original meaning, going to change it to the following:

> Docker supports a special [`.dockerignore`][dockerignore] file, which excludes files from the image build context based on file patterns.
> For example, the pattern `**/*.tmp` will ignore any files with the `.tmp` extension at the root build context directory and any of its subdirectories, recursively.
>
> You might be familiar with `.gitignore`, which excludes files from being committed to your repo.
> A common misconception is that `.dockerignore` is **additive** with `.gitignore`, but Docker does NOT use `.gitignore`!
> (Additionally, while they look very similar, the file patterns syntaxes are different.)

[dockerignore]: https://docs.docker.com/engine/reference/builder/#dockerignore-file

milas: The important points being:
 * Docker does NOT use `.gitignore` at all - I've seen confusion around this where people think that `.dockerignore` is "in addition to"
 * There are (not-so-)subtle differences in glob patterns between `.dockerignore` and `.gitignore`

ellenkorbes (approved):

PR #931, opened 2021-09-01
blog: Tilt + Rancher Desktop teaser

milas:
Very quick post on writing a `kim_build` function to use [kim][]
to build images for usage with [Rancher Desktop][].

[kim]: https://github.com/rancher/kim
[Rancher Desktop]: https://rancherdesktop.io/


ellenkorbes:
LGTM but could you update the title? I don't think anyone will be switching from 'Desktop Desktop' 😅

And what's with the Cloud Native Bingo stuff?

ellenkorbes (approved):

milas:
> LGTM but could you update the title? I don't think anyone will be switching from 'Desktop Desktop' 😅

omg 🤦 

> And what's with the Cloud Native Bingo stuff?

A joke that fell flat! Edited it to:
```
If your eyes glazed over during the second half of that sentence: images are built directly
_within_ the Kubernetes cluster using the same underlying technology (BuildKit) as Docker.
```

Both fixed in 2536e11. Thanks for review!

PR #945, opened 2021-09-24
api: generate builtin documentation for the new API

nicks:
Hello @milas, @landism,

Please review the following commits I made in branch nicks/codegen:

8942cabaf5ae73c0076aa48a550be75cb517d4b7 (2021-09-24 17:20:18 -0400)
api: generate builtin documentation for the new API


nicks:
in the medium-term, I think we'll want to break this up a bit and do some light contextual/navigational work, but for now just doing a dump to see how people react.

milas (approved):
Since they all start with `v1alpha1` and we're pretty good about linking to specific types in method docs, it's hopefully not too confusing for now 🤞

PR #949, opened 2021-09-29
docs: add in-section nav for tutorial

milas:
From the nav, sections can have "in-section" nav which adds
previous and next links (based on sidebar order).

For now, this will just be used for the tutorial.

in-section nav example


hyu (approved):
Nice touch!

PR #952, opened 2021-10-05
docs: add `tilt demo` to tutorial

milas:
This replaces most of the setup page and introduces some tweaks
to the `tilt up` step (which is, unfortunately and ironically,
actually a `tilt demo` step now).

There's a few other formatting/styling/typo tweaks I noticed while
proofing this.


hyu (approved):
Nice way to streamline this whole tutorial step! I think it'll be super interesting to see how users react to it, too, since we're now giving them more obvious guardrails.

PR #954, opened 2021-10-08
docs: Revamp UX

hyu:
Notes:

1. Since many typographic styles for blog and docs were shared, it was tricky to only make changes to the typography for docs without affecting the blog. I did my best, and erred on the side of simplifying the CSS so future changes will be easier. In some instances, I needed to make small tweaks to layout elements in the Blog to make things look roughly correctly after the changes to body and heading fonts.

2. I did a bit of work separating out CSS from the giant layout.scss into separate blog.scss and docs.scss files. This cleanup also included moving some styles to utility.scss and deleting styles that seemed unused.

3. The shared style variables for basic things like color and spacing are in a poor state. e.g., a new set of color variables were created for the homepage refresh a while back, but they co-exist with the old set of colors which are still widely used. Naming is confusing, too, with some styles prefixed with obscure stuff like "tilt2." In general, I used newer styles over older ones. (And also prevented myself from refactoring these shared variables let this PR balloon even more. 😬)

3. The way that navigation links are organized is based on mocks from Surbhi, but there are differences from the mocks due to recent changes (Milas's tutorial being merged in, removal of Live Update docs). I also exercised some judgment when there was a discrepancy between the mocks and the reality of what pages exist.

4. The link colors are now a blue that's closer to standard link colors. Visited links are grey like the surrounding text, but distinguised with an underline. This will hopefully make the text more readable than using the bright red links we had previously.

### Known Issues
a. Tiltfile API should be broken down into sections. It's currently a massive scroll! But, this is a bit involved since we use Sphinx to generate the content, and would need to use Sphinx to incorporate a TOC into the sidebar.

b. The new Tilt API is a separate page with a different layout for sidebar and content. It's very jarring! I need to make a separate PR for the [api repo](https://github.com/tilt-dev/api.tilt.dev/) to make the UX more consistent.

c. On mobile Docs, the top-level nav links go to special mobile-only navigation pages. (Basically, a page with just nav links and no content.) It would look closer to Surbhi's mocks if these nav pages have no header and footer, but that can be taken care of in a separate PR.

d. It'd be nice to go through all the docs pages and make sure the actual title is consistent with the title in the sidebar.


src/_includes/head.html
```diff
@@ -21,7 +21,11 @@
 
   <!-- tilt.dev typekit project -->
   <link rel="stylesheet" href="https://use.typekit.net/yii5fqs.css">
-  <link href="https://fonts.googleapis.com/css2?family=Fira+Code&display=swap" rel="stylesheet">
+
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+  <link href="https://fonts.googleapis.com/css2?family=Fira+Code:wght@300&display=swap" rel="stylesheet">
```
hyu, line 27: I'm specifying the exact weight of Fira Code we want to use. Also, I guess Google Fonts updated the boilerplate they ask you to copy. Supposedly this results in faster perceived loading time?

src/_layouts/docs.html
```diff
@@ -2,89 +2,80 @@
...
+        {% for link in site.data.docs.top %}
+          <li>
+            <a class="Docs-nav-link {% if page_href == link.href %}is-active{% endif %}" href="/{{link.href | escape}}">{{link.title | escape}}</a>
+          </li>
+        {% endfor %}
+      </ul>
+    </nav>
     <form class="js-only" action="/search" method="GET">
-      <div class="formItem">
-        <input type="text" name="q" id="q" placeholder="Search" aria-label="Search Documentation" />
```
hyu, line 9: I did a listen of this page with Voiceover, and I noticed that it reads the placeholder "Search"...so the additional label of "Search Documentation" seemed redundant and added slightly to the read time.

src/_includes/head.html
milas: It might be good to include a heavier weight as well - I noticed recently that if you have `<strong><code>foo</code></strong>` that it won't get bolded:

screenshot of monospace font not being bolded when wrapped in a strong tag

(notice the parameter names are not bold)

hyu: Thanks, @milas! I included a bold version of Fira Code.

src/_data/docs.yml
```diff
@@ -1,121 +1,237 @@
...
-    - title: Upgrade
-      href: upgrade.html
-    - title: "Writing a Tiltfile"
-      href: tiltfile_authoring.html
-    - title: "Tilt's Control Loop"
-      href: controlloop.html
-  - title: Tutorial
-    inSectionNav: true
+    href: /index.html
+    hrefMobile: /docs_nav_gettingstarted.html
```
milas, line 4: This pattern miiiight result in some slight weirdness for desktop users who have their browser at a small enough width to get the "mobile" breakpoint nav if they then resize the browser. (Alternatively, might be possible on mobile if you rotate the device.)

That said, it's enough of an edge case that it's probably not worth worrying about a ton. We can always sprinkle in some JS in the future if needed.

src/_includes/header-docs.html
```diff
@@ -0,0 +1,51 @@
+{%- include google-tag-manager-body.html -%}
+
+<header class="siteHeader2 siteHeader--docs">
+  <div class="siteHeader-logoAndMenuButton">
+    {% if page.layout == 'landing' %}
+    <!-- For the "Guide To App Devs" - /welcome_to_tilt.html -->
```
milas, line 6: I wonder if we should eliminate this page/layout at some point - it doesn't get a huge amount of views and people don't tend to stay on it particularly long.

(That said, it's got some inbound link sources, so we'd probably need a redirect or alternative)

src/_layouts/docs.html
```diff
@@ -1,74 +1,83 @@
...
-    {% endfor %}
-    </div>  
-
-  </nav>
-  <article class="docsContent">
-    <div class="docsHeader">
+          {% assign pages = section.items %}
+            <dt class="js-accordion__header Docs-subnav-title" data-accordion-opened="true">
+                {{ section.title | escape }}
+                {% svg assets/svg/carrot-down.svg %}
```
milas, line 51: FYI clicking ON the caret icon isn't triggering an accordion expand/collapse

src/_data/docs.yml
```diff
@@ -1,121 +1,237 @@
...
-      href: tutorial/3-tilt-ui.html
-    - title: 4. Code. Update. Repeat.
-      href: tutorial/4-code-update-repeat.html
-    - title: 5. Smart Rebuilds with Live Update
-      href: tutorial/5-live-update.html
-  - title: Example Projects
-    include: examples
-  - title: Onboarding Your Team
-    items:
+      href: index.html
```
milas, line 25: Weird Jekyll/Liquid thing, but right now the active page highlight doesn't work for this, if you make it `href: ''` it will. (It ends up seeing the default route as being an empty string rather than `index.html` which breaks the comparison)

src/_sass/about.scss
```diff
@@ -0,0 +1,18 @@
+.bio {
```
milas, line 1: This isn't specific to _this_ line of CSS but I noticed on the about page, the bullet indentation in the box is a bit strange now:

about page bullet point alignment

milas (approved):
WoOooOoo!

milas:
FYI I also did a quick pass across Chrome/Firefox/Safari (all on macOS) and didn't see any visual inconsistencies/oddities 👏

src/_data/docs.yml
hyu: Yes, I noticed this weirdness as well! And I agree that it seems like a bit of an edge case. I think some JS to finesse this layout resize behavior may be a great idea for a future PR.

src/_data/docs.yml
hyu: Ooh, great suggestion. Done!

src/_layouts/docs.html
hyu: Nice catch! Fixed.

src/_sass/about.scss
hyu: Oh yeah. By default, I made it so that <ul> has no spacing around it, which works better when we use <ul> for navigational elements. But it messed up layout on some random components. Thanks for catching this one! I fixed it, and will keep on the lookout for others that need help.

PR #977, opened 2021-11-09
docs: integrate https://github.com/tilt-dev/tilt-example-builders into the custom image builder guide

nicks:
Hello @chimney42,

Please review the following commits I made in branch nicks/builders:

335694a8d0284e08d4a79dc8d95be57472d90d56 (2021-11-09 15:12:34 -0500)
docs: integrate https://github.com/tilt-dev/tilt-example-builders into the custom image builder guide

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -819,7 +819,7 @@ def custom_build(ref: str, command: str, deps: List[str], tag: str = "", disable
       ['.'],
     )
 
-  Please read the `Custom Build Script How-to <custom_build.html>`_ on how to
+  Please read the `Custom Image Builders Guide <custom_build.html>`_ on how to
   use this function.
 
   All custom build scripts build an image and put it somewhere. But there are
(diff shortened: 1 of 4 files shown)


Chimney42 (approved):
LGTM.
Are there also docs that need to change in the tilt-extensions repo?

nicks:
I'm operating on the assumption that people don't really browse the extensions repo directly (and that assumption certainly seems consistent with the traffic numbers i've seen).

When people are writing a Tiltfile, they tend to go to the API reference (https://docs.tilt.dev/api.html) and search around in that (you can confirm this both from the analytics and from just asking people about it qualititatively).

When people are trying to solve a specific problem, they either learn about the solution from one of L's videos, or from searching for their problem (e.g., "remote image build"), and letting that lead them to the extension.

So i haven't been really trying to invest much in the tilt-extensions repo as a "main entrypoint" for docs, but rather assume it's kind of a docs leaf node

PR #978, opened 2021-11-09
docs: rewrite the guide to registries

nicks:
Hello @chimney42, @milas, @hyu,

Please review the following commits I made in branch nicks/registry:

96c8bcb6997c787205556a76626a90fc0423c57f (2021-11-09 16:45:26 -0500)
docs: rewrite the guide to registries

### DIFF
--- a/docs/personal_registry.md
+++ b/docs/personal_registry.md
@@ -1,5 +1,5 @@
 ---
-title: Choosing an Image Registry
+title: Setting up any Image Registry
 description: "A guide to dev image registries and how to get your code into a cluster."
 layout: docs
 sidebar: guides
@@ -23,11 +23,12 @@ For almost all clusters, the cluster will have a registry for you.
 - If you're using Docker for Desktop, there's no registry at all. You build
   directly into the container runtime.
 
-- If you're using Kind, [our setup scripts](choosing_clusters.html) will set up
-  a registry for you.
+- If you're using a local cluster like Kind, [our setup
+  scripts](choosing_clusters.html) will set up a registry for you.
 
 - If you're using a remote cluster (like AKS, EKS, GKE, and DigitalOcean
-  Kubernetes), you'll have a registry that's colocated with your cluster.
+  Kubernetes), you'll have a registry that's colocated with your cluster
+  and well-configured.
 
 ### Your Registry Secrets are Usually Already Set Up For You
 
@@ -40,7 +41,7 @@ Tilt never talks directly to a registry. When Tilt builds images, it tells the
 image builder to push to the registry when it's done. If you need to login to a
 registry, you'll login with the image builder.
 
-For Docker Hub, you run:
+For [Docker Hub](https://docs.docker.com/engine/reference/commandline/login/), you run:
 
 ```
 docker login
--- a/docs/personal_registry.md
+++ b/docs/personal_registry.md
@@ -1,27 +1,111 @@
 ---
-title: Setting a Personal Registry with a shared Tiltfile
-description: "Tilt's 'default_registry' function lets you change the image repository so you don't have to ask permission to get started."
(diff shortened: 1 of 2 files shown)


nicks:
Trying to accomplish a few things with this rewrite:
- orient the doc more towards beginning users, and clarify to them that they usually shouldn't have to think that hard about registries
- plug ttl.sh as a way to get started quickly when you don't have a registry
- set context correctly that the advanced stuff is really only for advanced users

open question: should we update tilt-avatars and other example projects to use ttl.sh by default? Then the examples would all work out of the box with remote clusters.

docs/personal_registry.md
```diff
@@ -1,27 +1,111 @@
...
+description: "A guide to dev image registries and how to get your code into a cluster."
 layout: docs
 sidebar: guides
 ---
 
-You should be able to start an app on Kubernetes from just the source code. One stumbling block is when YAML and scripts hard-code the image repository (the `gcr.io/windmill` in `gcr.io/windmill/user-service`). Tilt's `default_registry` function lets you change the image repository so you don't have to ask permission to get started. This guide walks you through two cases:
-* for users, how to experiment with an existing Tilt project by adding one line to the Tiltfile
-* for maintainers, how to configure your project so users don't need to modify the Tiltfile at all
+If you're building container images in dev, you'll need a place to put those
+images where your cluster can pull them.
```
Chimney42, line 9:
```suggestion
images from which your cluster can pull them.
```

docs/personal_registry.md
```diff
@@ -1,27 +1,111 @@
...
+If you're using an authenticated registry, that's usually not a problem either.
+Tilt never talks directly to a registry. When Tilt builds images, it tells the
+image builder to push to the registry when it's done. If you need to login to a
+registry, you'll login with the image builder.
+
+For Docker Hub, you run:
+
+```
+docker login
 ```
```
Chimney42, line 47: Suggest to add a link to the docs for `docker login` in case the user runs into any issues
https://docs.docker.com/engine/reference/commandline/login/

src/_data/docs.yml
```diff
@@ -87,7 +87,7 @@ sidebar:
       items:
         - title: Overview
           href: dependent_images.html
-        - title: Personal Registry
+        - title: Choosing an Image Registry
```
Chimney42, line 90: Imo this guide is not quite about choosing the registry, since in some cases that choice has already been made for you.
I would change the title to something like "Setting up any Image Registry with Tilt"

Chimney42 (approved):
Really great rewrite! This seems much more helpful now. 

Approved with some comments

docs/personal_registry.md
```diff
@@ -1,27 +1,111 @@
...
 
-## Experiment with an existing Tilt project
-You've cloned a repo, run `tilt up`, and see push errors saying you lack permission. Pick a registry that you can push to, like `gcr.io/my-personal-project`. Add the following line to the `Tiltfile`:
+## The Easy Way (for 95% of Users)
 
-```python
-default_registry('gcr.io/my-personal-project')
+Here's how you set up a dev image registry.
+
+Step 1) [Choose a cluster](choosing_clusters.html).
```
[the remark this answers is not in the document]
nicks: that's not what i mean though! even remote clusters usually have a image registry attached and auto-configured. (I've run into this in the past where people thought they had to create their own registry, and spent a lot of time fiddling with pull secrets until they realized their managed k8s service already has one setup)

src/_data/docs.yml
nicks: :+1:

docs/personal_registry.md
nicks: :+1:

PR #991, opened 2021-11-29
Add dc_resource/auto_init docs

nicksieger:
To be merged when tilt-dev/tilt#5225 is released.

- Update dc_resource docs to include auto_init
- Updated stars

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -388,6 +388,7 @@ TRIGGER_MODE_AUTO = type('_sentinel', (TriggerMode,),
 def dc_resource(name: str,
                 trigger_mode: TriggerMode = TRIGGER_MODE_AUTO,
                 resource_deps: List[str] = [],
+                auto_init: bool = True,
                 links: Union[str, Link, List[Union[str, Link]]] = [],
                 labels: Union[str, List[str]] = []) -> None:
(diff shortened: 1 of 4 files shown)


api/api.py
```diff
@@ -388,6 +388,7 @@ def trigger_mode(trigger_mode: TriggerMode):
 def dc_resource(name: str,
                 trigger_mode: TriggerMode = TRIGGER_MODE_AUTO,
                 resource_deps: List[str] = [],
+                auto_init: bool = True,
```
nicks, line 391: i think order matters, so this has to be after `labels`

nicks (approved):

api/api.py
nicksieger: I was mirroring `k8s_resource` and `local_resource` ordering. So you're saying that newly added named args need to be added at the end of the list, for compatibility?

nicks: yep, you did it correctly here: https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/docker_compose.go#L124

nicks: (we've talked to Alan + the starlark folks in the past about having an arg Unpacker that only allows keyword arguments, so that order doesn't matter: https://github.com/google/starlark-go/issues/249)

PR #997, opened 2021-12-09
Update listdir() documentation

remram44:
There are important differences with Python which I think are worth highlighting.

* Subdirectory names are not returned, only file names
* This function returns absolute paths, not relative to the input like Python
* Also fix typo 'rescursive'


nicks (approved):
thanks!

PR #999, opened 2021-12-10
docs: increase the max size of the accordion

nicks:
Hello @nicks,

Please review the following commits I made in branch nicks/accordian:

6dadc5ba4607a71bef844ee7a1f92cd2c5b2f67b (2021-12-10 18:09:10 -0500)
docs: increase the max size of the accordion


nicks:
cc @hyu not sure if there's a better way to fix this, but will at least make the docs less broken for now

PR #1004, opened 2021-12-13
docs: update for `docker_compose(Blob)`

nicksieger:
To be merged with 0.23.4 release

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -239,13 +239,15 @@ def docker_build(ref: str,
   """
   pass
 
-def docker_compose(configPaths: Union[str, List[str]]) -> None:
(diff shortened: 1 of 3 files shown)


lizzthabet (approved):
this looks great! the only thing that _might_ be confusing is the heading for the `docker_compose` section, which at a quick glance, makes me think that only paths are supported. (quickly) looking around at other heading entries, i didn't see an example of an either/or param, so it might not be worth worrying about. 

Screen Shot 2021-12-17 at 5 45 33 PM

nicksieger:
Yeah, I thought about that a bit. The code still needs to use that name for compatibility. The docs could change, I opted for consistency. Maybe someone will notice and then we can change it.

PR #1005, opened 2021-12-14
docs: Improve navigation for Tiltfile API section

Chimney42:
Please review the following changes for the Tiltfile API section:

- Changed order of content from most to least used (assumed)
- Added navigation items to directly point to functions, types, data and extensions


nicksieger:
yas
🎉

hyu (approved):
So much more sensible!

PR #1006, opened 2021-12-14
docs: improve landingpage intro

Chimney42:
Please review the following changes for the docs landing page:

- removed installation guide because I felt it didn't fit into the flow of the landingpage and there is a more extensive installation guide easily accessible from the landingpage through the navigation
- moved the "Run Tilt" section up above the explanation video
- improved wording of the "Run Tilt" section to guide users to the right manual according to their use case

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,5 +1,5 @@
 ---
-title: Tilt User Guide
+title: Getting Started With Tilt
 layout: docs
 sidebar: gettingstarted
 ---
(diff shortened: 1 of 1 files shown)


nicks (approved):
I'm not sure I'm a fan of removing the install script. I understand your argument in a vacuum. But there's such a strong social convention of having a copy-and-paste install script front and center on devtools sites, that i think people will be very annoyed if they can't find it on the main page.

e.g.
https://brew.sh/
https://helm.sh/
https://kind.sigs.k8s.io/

hyu (approved):
Nice! Hope this helps people find what they need a bit easier.

PR #1007, opened 2021-12-15
docs: Improve sidebar content for Tutorials

Chimney42:
Please review the following changes for the Tiltfile Tutorials:

- Change the title and sidebar content for Tutorial (aka new tutorial)
- Change the title and sidebar content for Tiltfile Guide (aka old tutorial)

PR #1008, opened 2021-12-15
docs: Improve access to language examples from sidebar

Chimney42:
Please review the following changes for the Guides section:

- added new "Migrating existing projects" section to make language examples more visible and remove unwanted association with the LiveUpdate feature

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,5 +1,5 @@
 ---
-title: Tilt User Guide
+title: Getting Started With Tilt
 layout: docs
 sidebar: gettingstarted
 ---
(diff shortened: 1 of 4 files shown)


nicksieger:
I think this also has commits from #1007, otherwise LGTM 👍

Chimney42:
> I think this also has commits from #1007, otherwise LGTM 👍

Yes this was intentional

hyu (approved):
These had been buried under the LiveUpdate guides, but I like the idea of surfacing these guides so people might be intrigued + look at examples.

Issue #1003, opened 2021-12-13
docs: unclear points about resource readiness checks


ahmetb:
Hi dear docs maintainers, I was reading https://docs.tilt.dev/resource_dependencies.html#adding-readiness-checks-for-startup-waiting and two prominent points made me think.

1. How is readiness defined? I think it is alluding to the Kubernetes status conditions described at https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties in other words, does Tilt look for:

	  ```
	  status:
	    conditions:
	    - type: Ready
	      status: "True"
	  ```

     The documentation does not make this clear.

2. Which type of workload objects does the Tilt look for, to check readiness? I see that the Pods are recognized (and presumably builtin workload controllers as well). However, for example:
    * does it know that it should wait for a Kubernetes Job to complete (a type that does not have a `Ready` condition) or,
    * how does a CRD that provides `Ready` condition (such as a knative Service CRD) work? do I need to opt-in explicitly as described in https://docs.tilt.dev/custom_resource.html#does-it-create-pods

I appreciate the answers. (But more importantly than answering to me, I wanted to point out that the docs do not make these points clear and I assume others wonder about similar things as they read the doc set.)

lizzthabet:
@ahmetb, those are great questions! i appreciate you opening an issue to ask for better documentation.

tilt looks at pods and jobs to determine readiness for kubernetes resources. a pod is considered ready when all of its containers are ready, (which may take into account readiness probes if they're defined on a container). a job is considered ready when it has completed. 

if a k8s workload object, including a CRD, doesn't create pods, there isn't a way for tilt to know when it's ready. that's why there's a `pod_readiness='ignore'` option, which tells tilt that it shouldn't be waiting for pods to show up to determine a resource's readiness. 

there are a couple [CRD examples](https://docs.tilt.dev/custom_resource.html#custom-resource-examples) linked in our docs, if that's helpful, or you can take a look at the [knative extension](https://github.com/tilt-dev/tilt-extensions/tree/master/knative) for more info about what that looks like in practice.

i'll follow up with a docs pr, too.

ahmetb:
> if a k8s workload object, including a CRD, doesn't create pods, there isn't a way for tilt to know when it's ready.

[...]

PR #1009, opened 2021-12-15
docs: add better detail on Kubernetes resource readiness checks

lizzthabet:
Add more details to how Tilt determines readiness on k8s resources, based on feedback from #1003.

### DIFF
--- a/docs/resource_dependencies.md
+++ b/docs/resource_dependencies.md
@@ -81,7 +81,7 @@ Once a resource is ready, Tilt will start building the resources that depend on 
 
 By default, a resource is "ready" when:
 
-- For `k8s_resource`: the pod is running and Kubernetes considers all of its containers ready.
+- For `k8s_resource`: the pod is running and Kubernetes considers all of its containers ready. A job is considered ready when it has completed.
 
 - For `dc_resource`: the container is started (NB: Tilt doesn't currently observe docker-compose health checks).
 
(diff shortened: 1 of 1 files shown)


docs/resource_dependencies.md
```diff
@@ -96,6 +96,15 @@ Kubernetes has a built-in notion of
...
 Tilt will use the built-in readiness probe when it's available.
 
+For any Kubernetes object type, including built-ins and CRDs, that creates pods,
+Tilt will consider it ready when the pod is running and all of its containers
+are ready. The exception to this rule is jobs, which are considered ready by Tilt when
+the job has completed.
+
+If an object does not create pods, there isn't a way for Tilt to determine readiness.
+Instead, you'll want to configure the resource to let Tilt know that it shouldn't
+wait to see pods to determine readiness.
```
milas, line 106:
```suggestion
Instead, you'll want to configure the resource to let Tilt know that it shouldn't
try to use pods to determine readiness. Resources configured to ignore pods
for readiness will appear as ready in the Tilt UI as soon as they are applied to
the Kubernetes cluster.
```

This explanation is great!

You don't have to take my suggestion as-is, but I think it might be useful to have another sentence explaining what will happen after making this change.

milas (approved):

docs/resource_dependencies.md
Chimney42: Could you link to documentation that explains how I can ignore pods for readiness, here?

Chimney42 (approved):
🎉

docs/resource_dependencies.md
lizzthabet: sounds good! thank you both. 

there's a code example right below this paragraph, but i'll also link to the api docs so it's super clear.

PR #1016, opened 2021-12-21
docs:  Make it possible to navigate to individual functions in Tiltfile reference from sidebar

lianmakesthings:
https://app.shortcut.com/windmill/story/13135/make-it-possible-to-navigate-to-individual-functions-in-tiltfile-reference-from-sidebar

The Tiltfile API Reference contains subsections Functions, Types and Data. These subsections are rendered as their own accordions. If an item in one of the subsections is active, the accordion is opened and the item is highlighted

Hi @hyu @lizzthabet @milas, please review the following changes:

- Auto generate navigation items, so all sub items are in the navigation bar
- Render subsections as accordions, closed by default
- Highlight current subitem
- Auto open accordion if subitem is active

Tested on: MacOS Firefox & Safari, iPhone 13 Firefox & Safari


hack/gen_api_types_toc_helper.py
```diff
@@ -0,0 +1,36 @@
...
+
+with open('src/_data/docs.yml') as f:
+  lines = f.readlines()
+  start_line = find_index(lines, lambda l: '# Start Tiltfile Types Reference' in l)
+  end_line = find_index(lines, lambda l: '# End Tiltfile Types Reference' in l)
+  if start_line == -1 or end_line == -1:
+    exit(1)
+
+  new_lines = lines[0:start_line+1] + typeCmds + lines[end_line:]
+  print(''.join(new_lines).rstrip())
```
lizzthabet, line 36: it looks like these three files are mostly doing the same thing, minus the yaml file name and overview entry. do you think it's worth creating a single file for the api toc generators and adding a loop that goes through `functions`, `types`, and `data`?

(i can't say i've written that much python, so if you're in the same boat and think this might take too much work, that's ok!)

src/assets/js/subsection-highlighting.js
```diff
@@ -0,0 +1,27 @@
...
+    let activeItem = document.querySelector(".Docs-subnav-link.is-active");
+    if (activeItem && activeItem.href != targetUrl) {
+        activeItem.classList.remove("is-active");
+        setTimeout(activateTarget, 300);
+    } else {
+        activateTarget();
+    }
+}
+
+highlightSubnav(document.URL)
```
lizzthabet, line 27: i think it'd be good to wrap this call in an iife, just to make sure the timing is okay when it runs. you can follow the example of defining a listener and attaching it to the `readystatechange` document event in `link.js`. something like: 
```js
(function () {
  if (document.readyState !== "complete") {
    return
   }

  function eventListener () {
    highlightSubnav(document.URL)
    document.removeEventListener('readystatechange', eventListener)
  }

  document.addEventListener('readystatechange', eventListener);
})();

```

lizzthabet (approved):
this looks great! 🎉 

i will note for other testers that the `link.js` logic will erroneously add `target="_blank"` properties to all the api subsection links. this works correctly during local dev and (presumably) the live site, but not for the netlify previews. 😅 this isn't in the scope of this pr, so i'll put up a separate pr to add netlify to the "links in new windows" blocklist, and our previews are more accurate!

hack/gen_api_types_toc_helper.py
lianmakesthings: kinda, the way it works right now is that it prints the generated lines and in the bash script that output is being piped into docs.yml
If I put everything in one file, I would have to rewrite the code so I would generate the navigation items, and find the right placement for all sections "at once" which I was concerned would make it harder to read & understand.

src/assets/js/subsection-highlighting.js
lianmakesthings: ah yes, I was going to do that actually and it slipped my mind. Thanks for calling this out!

lizzthabet:
oops, i just noticed a bug! there's a section of the "functions" functions that are `#modules` and not `#api`. these sidebar links are broken from `os.getcwd` to `v1alpha1.ui_text_input_spec`. ([here's a link](https://github.com/tilt-dev/tilt.build/pull/1015/files#diff-0d441c57fa701e058508a0332f76bc86c7b020ecf3cc6b5f41b68eea8daa1dacR14-R16) to where that happens in `functions.md`.)

lizzthabet:
one more consideration: whenever i click on a link to an api entry _outside_ the sidebar nav, the active link class doesn't change. i don't think we need to deal with that in this pr, but it might be worth a follow-up.

(sorry for the approval and then extra comments as i test some more!)

hyu:
Thanks for doing this, @lianmakesthings! This is a great improvement for people who want to get a quick overview of the Tiltfile API.

The design that happens by default, though, when you have an expandable menu item inside another one looks a little off (e.g., Tiltfile API Reference > Functions)

image

The original design by Surbhi was to make the secondary nav have a white background, with children that are further indented:

image


I'm noting this and CC'ing @SurbhiSGupta for awareness. But, that said, I don't think that should hold up this PR, since the navigation is still functional and immediately becomes obvious once you interact with it.

hyu (approved):

src/assets/js/subsection-highlighting.js
lianmakesthings: This code won't work I think, because if the document is not ready we're returning and don't attach the eventlistener to the readystate.
I implemented this instead, it should do the trick, I think

```
(function () {
  function eventListener () {
    highlightSubnav(document.URL)
    document.removeEventListener('readystatechange', eventListener)
  }
  document.addEventListener('readystatechange', eventListener);
})();
```

lianmakesthings:
> one more consideration: whenever i click on a link to an api entry _outside_ the sidebar nav, the active link class doesn't change. i don't think we need to deal with that in this pr, but it might be worth a follow-up.
> 
> (sorry for the approval and then extra comments as i test some more!)

Could you create a separate issue for that so we keep track of it?

lianmakesthings:
> The original design by Surbhi was to make the secondary nav have a white background, with children that are further indented:
> 
> image
> 
> I'm noting this and CC'ing @SurbhiSGupta for awareness. But, that said, I don't think that should hold up this PR, since the navigation is still functional and immediately becomes obvious once you interact with it.
> 
> (That said, if you want to make a quick fix, I would just make the item's background white and change the caret color to dark gray.)

The children are already further intended and with a smaller fontsize, so I hope this is already clear for now.

I made the background white, but I personally feel having the caret in a dark gray looks out of place with the rest of the colors.
Screenshot 2021-12-23 at 14 44 48

I personally like the light gray better, but it might be too light considering accessibility. 
Screenshot 2021-12-23 at 14 45 11

@hyu @SurbhiSGupta what do you think?

PR #1017, opened 2021-12-23
docs: Change nav to prioritize tutorial

hyu:
@SurbhiSGupta asked to move some of the more "expert" articles to a section below the tutorial. 

2021-12-22 - Docs nav

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -30,7 +30,7 @@ iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.co
 
 ## Finding the right manual
 **Completely new to Tilt?**  
-Watch our two minute [explanation video](#watch-tilt-in-two-minutes).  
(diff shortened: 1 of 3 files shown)


lianmakesthings:
In this context, I feel the Title "Learn about Tilt" doesn't make much sense anymore. Maybe we can change it to "Discover Tilt"?
I was also thinking about "Quickstart" but then I feel we'd have to rename the "Quicklinks" section too because it would sound too similar 🤔  I'd like to think on this some more

lianmakesthings (approved):

lianmakesthings:
> In this context, I feel the Title "Learn about Tilt" doesn't make much sense anymore. Maybe we can change it to "Discover Tilt"? I was also thinking about "Quickstart" but then I feel we'd have to rename the "Quicklinks" section too because it would sound too similar 🤔 I'd like to think on this some more

Now that I've moved the "Is Tilt right for me" page up to the top section, I feel this works now.

PR #1020, opened 2022-01-02
Docs fixes as per design spec

SurbhiSGupta:
I've tried to implement some of the visual changes to our docs as discussed with @hyu. Here are the [notes](https://docs.google.com/document/d/1TKLlsvkvMzBFBN3nDD7s_qHMB25QXFbDSyt5j7qKhXI/edit) about visual feedback on docs to guide most of these changes.

# Changes Made

A: Left align logo
B: Reduce header height
C: Reduce space between nav and content
D: Change color of separating line
E: Create left column divider line and add max column width
F: Change color of carrot
G: Change padding around titles
H: Change weight of H1
I: Changed line color
J: Horizontal center aligned toolbar
K: Sticky primary nav

Screen Shot 2022-01-02 at 3 49 49 PM

## Testing on Desktop
image

## Testing on Medium width
image

## Testing on mobile
image

# What is not covered

1. Carrot icons and animation did not change on the left sidebar. When the titles expand or collapse, the icon remains stationary in 'expanded' position
image

2. H2 and H3 have an unpredictable color behavior. I have not touched that.
image

3. Search bar has no left/ right padding for medium sized screen
image

4. Footer tabs are not justified or spaced correctly
image

5. Right hand toolbar - 2 styles of buttons / links - spacing is strange
image

6. Search bar in desktop is small - it strangely stretches to full page width on medium sized screen
image

7. Footer for smaller screens - the footer requires a different color background and a visual treatment to make them look like tabs.
image


nicksieger:
Love this, looks fantastic. Also thanks for the detailed breakdown of changes!

src/_sass/docs.scss
```diff
@@ -21,13 +21,13 @@ $sidebar-maximum-width: 400px;
 .Docs-wrapper {
   display: flex;
   flex-direction: column;
-  margin-top: $spacing-unit;
+  margin-top: 0px;
```
hyu, line 24: You can remove `margin-top: 0px;` entirely and have the same layout, since `.Docs-wrapper` has no margin by default.

src/_sass/tilt.scss
```diff
@@ -74,7 +74,7 @@ $tilt2-font-weight-mono-bold: 600;
 
 $tilt2-font-weight-sans-light: 300;
 $tilt2-font-weight-sans-bold: 700;
-$tilt2-font-weight-sans-black: 900;
+$tilt2-font-weight-sans-black: 600;
```
hyu, line 77: This sans-black variable is used in the mixin `font-hero`, which is used for the big bold text on the homepage and enterprise page. So changing the value here has unintended effects.

A better way to do "H: Change weight of H1" might be to use the lighter weight font in docs.scss:

```
.Docs-content-title {
  h1 {
    font-weight: $tilt2-font-weight-sans-bold;
```

The assumption I made here is that we're restricting ourselves to just a few weights of the sans font, which is why I made just 3 weight variables. But, if you think it's a pattern that we want to use other weights, too, we might want to add another weight.

src/_sass/header.scss
```diff
@@ -1,13 +1,11 @@
...
-  margin-right: auto;
-  margin-left: auto;
+  margin-right: $spacing-unit;
+  margin-left: $spacing-unit;
 
   @include mobileAndAbove {
-    max-width: $tilt2-header-width;
-    padding-top: $spacing-unit * 2;
-    padding-bottom: $spacing-unit;
+    padding-top: $spacing-unit * 1;
```
hyu, line 7: These changes for the header also apply to the home page and blog, which I don't think is intended. 

I started work already in a separate branch which has a separate header for Docs vs homepage / blog. You can see that I made a separate [header-docs.scss](https://github.com/tilt-dev/tilt.build/blob/hyu/slim-header/src/_sass/header-docs.scss) that corresponds to the separate docs-header.html.

Maybe we can pair and integrate some of my changes with your branch!

hyu (changes requested):

src/_sass/docs.scss
```diff
@@ -164,6 +168,10 @@ $sidebar-maximum-width: 400px;
 // --------------------------------------------------
 .Docs-sidebar {
   background-color: $color-white;
+  border-right: 2px solid $tilt2-color-gray-lightest;
+  position: sticky;
+  top: 64px;
```
hyu, line 173: When I tested, this gap at the top of the sidebar is just a bit shorter than the height of the sticky horizontal menu with Docs categories. We should make these match up so the overall effect is more refined.

In the past, I've done this by using a shared variable to set the height of the menu _and_ the sidebar-top-gap.

src/assets/svg/carrot-down.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="24" height="14" viewBox="0 0 24 14" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path fill-rule="evenodd" clip-rule="evenodd" d="M2.59338 0.859505C2.00011 0.266233 1.03823 0.266233 0.444954 0.859504C-0.148318 1.45278 -0.148318 2.41466 0.444954 3.00793L10.8762 13.4392C11.0006 13.5636 11.1411 13.6618 11.2911 13.7341C11.8672 14.0471 12.603 13.96 13.0902 13.4728L23.555 3.00795C24.1483 2.41468 24.1483 1.45279 23.555 0.859522C22.9618 0.26625 21.9999 0.26625 21.4066 0.859522L12 10.2661L2.59338 0.859505Z" fill="white"/>
+<svg width="20" height="10" viewBox="0 0 24 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
hyu, line 1: In general, I prefer removing the `fill=` from SVGs so we can set the color using CSS. That makes it possible to re-use SVGs and give it different color treatments for hover states. Also, when we set color using CSS, we can use variables that are easier to read and update.

hyu:
@SurbhiSGupta, I'm so excited to see you make a PR with so many nice changes to improve the design of the Docs! This is a wonderful development. (Though, gotta give you the obligatory side-eye here about working over vacation 😒)

I'd love to hear about your experience working with the frontend code — there's probably real rool for improvement especially in the way the variables work at the moment.

Also, I blocked out some pairing time for Wed in case you do want to integrate the "separate Docs header" approach I mentioned.

PR #1023, opened 2022-01-03
docs: remove the tests doc, which was an experimental feature that never launched

nicks:
Hello @lianmakesthings,

Please review the following commits I made in branch nicks/test:

563551b0a43cb1b322728e94d0e9f20d507048b7 (2022-01-03 18:40:41 -0500)
docs: remove the tests doc, which was an experimental feature that never launched


lianmakesthings (approved):
Just to clarify. Is the function still in the code, though? If yes, we should remove that as well

nicks:
yup, i think that's right

PR #1026, opened 2022-01-06
docs: document List[str] possibility for all commands

landism:
1. Make sure that command parameters that accept `Union[str | List[str]]` are documented as such. (I went by calls to `ValueGroupToCmdHelper` and `ValueToUnixCmd`, but maybe there are others?)
2. Make the language around command/bat param interaction a bit more consistent.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -1058,7 +1058,7 @@ def allow_k8s_contexts(contexts: Union[str, List[str]]) -> None:
   pass
 
 def enable_feature(feature_name: str) -> None:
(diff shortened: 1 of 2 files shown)


api/api.py
```diff
@@ -1045,7 +1058,7 @@ def allow_k8s_contexts(contexts: Union[str, List[str]]) -> None:
   pass
 
 def enable_feature(feature_name: str) -> None:
-  """Configures Tilt to enable non-default features (e.g., experimental or deprecated).
```
nicks, line 1048: ?

nicks (approved):

PR #1027, opened 2022-01-06
docs: fix right margin in api code blocks

landism:
### Before
The negative margin extends the code block border over the next character (which is not in the code block), e.g.:
image
image

### After
image
image

I spent a bit reading docs and haven't managed to understand the underlying issue (why is it an issue on the right but not the left? does this affect other sphinx projects?) or why this solves it (most pages I found discussing inline vs inline block indicated only differences that don't apply here, though I did find lots of handwavy references to inline-block changing the way padding is calculated wrt parent elements). It seems to work and I'm guessing it's not worth more investigation at the moment.

### DIFF
--- a/src/_sass/api.scss
+++ b/src/_sass/api.scss
@@ -72,9 +72,3 @@ dl:not(.class) > dt {
 dt:hover .headerlink::after {
   opacity: 1;
 }
-
-// Sphinx adds spaces after every literal,
-// so add negative spacing to tighten up the text.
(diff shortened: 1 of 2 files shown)


nicks (approved):

hyu (approved):
Thanks also, @landism, for the fascinating side discussion troubleshooting this. TIL yet another weirdness of how whitespace is handled vis-a-vis box model

PR #1029, opened 2022-01-10
docs: static Snippets Library

nicksieger:


hyu:
Did a design pass!

Screen Shot 2022-01-21 at 18 00 00

src/_data/snippet_order.yml
```diff
@@ -0,0 +1,22 @@
+---
+# Snippets mentioned here will be rendered first in this order in snippets.md;
```
lianmakesthings, line 2: What's the purpose of keeping this specific order, and how do we determine which ones go first?

lianmakesthings (approved):

src/_data/snippet_order.yml
nicksieger: When I created the initial list I roughly ordered it with core/basic snippets early in the list. I started with snippets I got from "Tiltfile Concepts" and added a few more from some of the intro sections. Once we add search/tag/filter I don't think the ordering will be quite as important. If we want to maintain the ordering we can move snippets around to put whatever we want above the fold.

PR #1031, opened 2022-01-11
docs: extend docker_build docs

lianmakesthings:
Hi @nicks @milas,

could you please review the following changes I made to the docs:
- Added a bit in the docker_build section to inform user that this builtin will also push to an image registry in some instances

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -218,6 +218,8 @@ def docker_build(ref: str,
 
   When using Docker Compose, Tilt expects the image build to be either managed by your Docker Compose file (via the `image <https://docs.docker.com/compose/compose-file/compose-file-v3/#build>`_ key) OR by Tilt's :meth:`docker_build`, but not both. (Follow this `GitHub issue <https://github.com/tilt-dev/tilt/issues/5196>`_ to be notified of changes to this expectation.)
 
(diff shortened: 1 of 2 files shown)


api/api.py
```diff
@@ -218,6 +218,8 @@ def docker_build(ref: str,
 
   When using Docker Compose, Tilt expects the image build to be either managed by your Docker Compose file (via the `image <https://docs.docker.com/compose/compose-file/compose-file-v3/#build>`_ key) OR by Tilt's :meth:`docker_build`, but not both. (Follow this `GitHub issue <https://github.com/tilt-dev/tilt/issues/5196>`_ to be notified of changes to this expectation.)
 
+  Finally, Tilt will push the built image to a local image registry if one exists or to docker.io by default. Tilt can detect if you are using Docker for Desktop or Minikube and omit pushing.
```
nicks, line 221: This is not really accurate in a few ways. 

For example, the "docker.io by default" part is kind of misleading...if you have a in image named 'golang', that image is stored at https://hub.docker.com/_/golang and pushed/pulled from the url docker.io/library/golang.

the problem here isn't your fault...it's that the image distribution specs are kind of mess of disagreements and misinterpretations.

I might say something here to the effect of 

> Finally, Tilt will put the image in a place where the target runtime can access it. Tilt will make a best effort to detect what kind of runtime you're using (Docker Compose, Kind, GKE, etc), and pick the best strategy for getting the image into the cluster fast. See https://docs.tilt.dev/choosing_clusters.html for more info.

PR #1036, opened 2022-01-13
docs: add support for tags in docs menu and page titles

lizzthabet:
This PR adds support for adding a `pilltag` to a doc menu item and page heading. The only tag that's actually added to the live docs is the `beta` tag for the Disable Resources docs. I've removed the `new` tag next to the "Tilt API" header link.

Here are a couple screenshots of what that looks like with tags added to more places:

Screen Shot 2022-01-13 at 4 19 03 PM

Screen Shot 2022-01-13 at 4 24 11 PM

### DIFF
--- a/docs/disable_resources.md
+++ b/docs/disable_resources.md
@@ -3,9 +3,10 @@ title: Disabling Resources Through Tilt
 description: "Use Tilt's new feature to disable and enable resources through the UI. Manage what resources you have up and running more seamlessly."
 layout: docs
 sidebar: guides
+pilltag: beta
 ---
 
-> <span class="pill-tag">Beta</span> This feature is in beta. It's turned off by default, and breaking changes could be introduced while under development.
+> 💡 This feature is in beta. It's turned off by default, and breaking changes could be introduced while under development.
 
 Through Tilt’s UI Dashboard, you can enable and disable resources that are available in your Tilt session.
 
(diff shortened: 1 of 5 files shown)


lianmakesthings:
@SurbhiSGupta do you have any thoughts on the pill design with the longer text?

SurbhiSGupta:
Keeping in mind, the main purpose of tags, (discoverability by drawing attention), I think it would be good to show tags only for beta (and maybe deprecated - inline with content not sidebar)

That would basically self-solve the problem. But _just in case_, here are some ways of showing the rest of the tags anyways:

1. Tags with truncated names that gives information on hover
     ` Beta `, ` Exp… `, ` Rel… `, ` Dep `

2. Emojis with hover tooltip
     ✨ = Beta or new to see
     🧪 = Experimental

3. Icons with key legend a the bottom of the list

lizzthabet:
> Keeping in mind, the main purpose of tags, (discoverability by drawing attention), I think it would be good to show tags only for beta (and maybe deprecated - inline with content not sidebar)

that makes sense to me! i think the `beta` and `new` tags are what we'd most likely want to call attention to in the navigation. other tags can be included in their page's (sub)heading, which is supported now. (i don't believe there's template support for tagging api doc entries, but that'd be easy enough to add if/when we'd like to.)

any objections to merging this as-is?

SurbhiSGupta:
Lgtm as is !

lianmakesthings (approved):

PR #1038, opened 2022-01-18
docs: update the helm docs to recomment helm_resource()

nicks:
Hello @lianmakesthings, @nicksieger,

Please review the following commits I made in branch nicks/helmresource:

a4656c319235615a6382327ec351191b56ef855a (2022-01-18 15:22:41 -0500)
docs: update the helm docs to recomment helm_resource()

### DIFF
--- a/docs/helm.md
+++ b/docs/helm.md
@@ -7,51 +7,74 @@ sidebar: guides
 
 Tilt supports Helm out-of-the-box.
 
-There are two ways you can use Helm charts with Tilt.
+There are two major ways we see teams use Helm charts with Tilt.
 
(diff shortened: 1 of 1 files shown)


nicksieger (approved):
Changes look great overall. 

So is there still a use case for `helm_remote`? Iterating on a remote chart, basically, or just evaluating the remote chart as a template?

nicks:
`helm_remote` has a lot of existing users. So it's fairly battle-hardened and will work for more cases in the short term. But I expect that the approach `helm_resource` takes is going to be more robust long-term.

lianmakesthings (approved):

PR #1039, opened 2022-01-20
docs: adds entry to docs for color extension

lianmakesthings:
Hi @milas 

- adds a new entry for color extension to docs


milas (approved):
Thanks for doing this! 😻

PR #1040, opened 2022-01-21
docs: Remove Welcome To Tilt page

lianmakesthings:
Hi @lizzthabet @SurbhiSGupta 

- removes all references to the "Welcome to Tilt" page
- removes the "Welcome to Tilt" page


src/_includes/header-docs.html
```diff
@@ -3,7 +3,7 @@
 <header class="siteHeader2 siteHeader--docs">
   <div class="siteHeader-logoAndMenuButton">
     {% if page.layout == 'landing' %}
-    <!-- For the "Guide To App Devs" - /welcome_to_tilt.html -->
```
lianmakesthings, line 6: I just removed this, so I can reference it in the PR 😅 
@lizzthabet what's the significance of this comment? Can we remove the following <a> tag as it won't be needed anywhere else?

SurbhiSGupta (approved):

src/_includes/header-docs.html
hyu: @lianmakesthings, thanks for noticing this defunct code. I made a separate [PR](https://github.com/tilt-dev/tilt.build/pull/1042) to clean this up, among other things.

lizzthabet: thanks, @hyu and @lianmakesthings! i'm not very familiar with the history of that landing page, but i know i've been directed to it from other places before. this is all i can dig up atm: https://github.com/search?q=welcome_to_tilt+user%3Atilt-dev&type=code.

lizzthabet: ...which as i'm looking through your pr, you've already handled! apologies for the unhelpful/premature comment. 😅 i can't see any reason to keep the layout block for a page we're not using anymore.

lizzthabet (approved):

hyu (approved):
Thanks for cleaning up this old page!

PR #1041, opened 2022-01-21
docs: update extension overview in docs

lianmakesthings:
Hi @nicks @nicksieger,

- updated docs to add all missing extensions

It looks like there were a lot of extensions missing. I'm wondering if there's an easier (=automated) way to do this. Maybe a CircleCI job that scrapes [the readme](https://github.com/tilt-dev/tilt-extensions/blob/master/README.md)?


nicksieger (approved):
I think this is great overall! I do wonder if we should de-list some of the extensions that are older that don't represent the way we recommend people to use Tilt. None of them stand out like that to me upon looking though. 🤷‍♂️

nicks (approved):

PR #1045, opened 2022-01-25
docs: add titles to snippet reference links

milas:
Change the YAML structure so links have both a title and href:
```
link:
  title: "foo"
  href: "/foo.html#bar"
```

Each snippet now shows a "References" footer with the link.

### With Reference Link
Screenshot of snippet with references footer including link

### Without Reference Link
Screenshot of snippet without references footer


nicksieger (approved):
Great!

PR #1048, opened 2022-01-26
add readme for contributing snippets

lianmakesthings:
Hi @hyu @SurbhiSGupta @nicksieger,

- adds a Readme, explaining how to contribute a snippet
- reference Readme in snippet library
- fixed styling, so the top content is not hidden behind header when navigating to anchor links
- moved copy-to-clipboard button away from right edge

I specifically did not include `tags` and `release_stage` in the template, since they are currently only for internal usage.

@SurbhiSGupta I also included an example design for adding a contributor's handle to a snippet. I'm 0% committed to the design, though. Interested to hear your suggestions!
Screenshot 2022-01-26 at 21 19 42

### DIFF
--- a/contributing-snippets.md
+++ b/contributing-snippets.md
@@ -18,4 +18,4 @@ contributor: Your Github Handle (optional)
 3. Commit your changes and submit a Pull Request
 
 4. A member of the Tilt team will review your snippet and modify it if necessary.  
-We might decide to not include it in our library. If that's the case, you will receive an explanation.
+If there's a reason that we don't think it's right for our Snippets Library, we'll let you know why. We appreciate you taking the time to contribute!
--- a/contributing-snippets.md
+++ b/contributing-snippets.md
@@ -1,0 +1,19 @@
+Do you have a suggestion for a Tiltfile snippet?
+Submitting a snippet is as straightfoward as submitting a Pull Request.
+
+0. Clone this repository
+
+1. Create a file in src/_data/snippets/*snippet-name*.yml
+
+2. Copy the following content and modify as applicable
+```
+title: Descriptive Name of Your Snippet
+description: Explain what your snippet does
+code: |
+  *Your Snippet Goes Here*
+contributor: Your Github Handle (optional)
+```
+
+3. Commit your snippet and [submit a Pull Request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request)
+
+4. A member of the Tilt team will review your snippet and modify it if necessary. We might decide to not include it in our library. If that's the case, you will receive an explanation. 
--- a/contributing-snippets.md
+++ b/contributing-snippets.md
@@ -1,4 +1,5 @@
-Do you have a suggestion for a Tiltfile snippet?
+# Contributing to Tiltfile Snippets
+Would you like to help improve our Snippet Library?  
 Submitting a snippet is as straightfoward as submitting a Pull Request.
 
 0. Clone this repository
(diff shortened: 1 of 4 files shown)


contributing-snippets.md
```diff
@@ -0,0 +1,21 @@
+# Contributing to Tiltfile Snippets
```
nicksieger, line 1: We don't have a regular CONTRIBUTING.md file in this repo, I wonder if this should be it and we add a little bit of boilerplate for contributing documentation fixes?

contributing-snippets.md
```diff
@@ -0,0 +1,21 @@
...
+Would you like to help improve our Snippet Library?  
+Submitting a snippet is as straightfoward as submitting a Pull Request.
+
+0. Clone this repository
+
+1. Create a file in src/_data/snippets/*snippet-name*.yml
```
nicksieger, line 7: Changed to snake-case to match the current convention of files in the directory
```suggestion
1. Create a file in src/_data/snippets/*snippet_name*.yml
```

contributing-snippets.md
```diff
@@ -0,0 +1,21 @@
+# Contributing to Tiltfile Snippets
+Would you like to help improve our Snippet Library?  
+Submitting a snippet is as straightfoward as submitting a Pull Request.
+
+0. Clone this repository
```
nicksieger, line 5:
```suggestion
0. Fork this repository on Github and then clone from your forked version of this repository
```

docs/snippets.md
```diff
@@ -5,6 +5,9 @@ sidebar: guides
 hideHelpfulForm: true
 ---
 
+> **Want to contribute?**  
+> Follow [this guide](https://github.com/tilt-dev/tilt.build#contributing-snippets) to find out how to your own snippets!
```
nicksieger, line 9: I think this will need to be
```suggestion
> Follow [this guide](https://github.com/tilt-dev/tilt.build/blob/master/contributing-snippets.md) to find out how to your own snippets!
```

contributing-snippets.md
lianmakesthings: Good point.
I do wonder though how much this will blow up the scope of this particular story (given that I'm also going to be gone for two weeks)
Here's my suggestion: For this story, leave it as is. 

Follow up afterwards in a separate PR to make this a more generic contributing md?

SurbhiSGupta: Design looks simple and good, however, I would recommend treating color of 'user name' slightly different for easier readability. Would it make sense to include a 'date' stamp for when the snippet was submitted?

Just to recap, for each snippet the following information will be included:

- Title
- 1 to 2 line Context
- Actual snippet
- Copy icon
- Permalink
- Link to the docs (described better)
- _(maybe)_ Submitted by `user name` 
- _(maybe)_ Tags

lianmakesthings: Personally, I don't feel like a submittal date would add any relevant information and risk visually overloading the design.

nicksieger (approved):

lianmakesthings:
@SurbhiSGupta I aligned the contributor and permalink like this. Wdyt?

Screenshot 2022-01-27 at 20 52 37

nicksieger:
A minor point -- I noticed we have a chain-link icon elsewhere when hovering over a heading to point to a fragment in a page. Should we use that instead of "Permalink"?

lianmakesthings:
> A minor point -- I noticed we have a chain-link icon elsewhere when hovering over a heading to point to a fragment in a page. Should we use that instead of "Permalink"?

Ha! @SurbhiSGupta and myself were discussing this exact topic yesterday! Seems good to keep things consistent, I just assumed you made a conscious design decision to put in the `permalink` so it stands out. cc'ing @hyu into this convo to see what they think.
The chainlink thingie is done via CSS afaics https://github.com/tilt-dev/tilt.build/blob/master/src/_sass/docs.scss#L323

contributing-snippets.md
```diff
@@ -0,0 +1,21 @@
...
+title: Descriptive Name of Your Snippet
+description: Explain what your snippet does
+code: |
+  *Your code goes here*
+contributor: Your Github Handle (optional)
+```
+
+3. Commit your changes and submit a Pull Request
+
+4. A member of the Tilt team will review your snippet and modify it if necessary.
```
hyu, line 20: What about this edit? To me, this seems more respectful that people are taking time out of their busy schedules to even submit a snippet, and expresses more appreciation and respect for that time.


> A member of the Tilt team will review your snippet and suggest edits if necessary. If there's a reason that we don't think it's right for our Snippets Library, we'll let you know why. We appreciate you taking the time to contribute!

hyu (approved):
Thanks for writing this, @lianmakesthings! Very straightforward and clear.

contributing-snippets.md
```diff
@@ -0,0 +1,21 @@
...
+description: Explain what your snippet does
+code: |
+  *Your code goes here*
+contributor: Your Github Handle (optional)
+```
+
+3. Commit your changes and submit a Pull Request
+
+4. A member of the Tilt team will review your snippet and modify it if necessary.  
+We might decide to not include it in our library. If that's the case, you will receive an explanation.
```
lianmakesthings, line 21:
```suggestion
If there's a reason that we don't think it's right for our Snippets Library, we'll let you know why. We appreciate you taking the time to contribute!
```

PR #1054, opened 2022-02-02
api: clear_enabled_resources

landism:


lizzthabet:
lgtm, minus a missing nav entry for the new method!

landism:
> lgtm, minus a missing nav entry for the new method!

oh no, we don't automatically generate that like we do the function list at the top of the page?

lizzthabet (approved):

PR #1060, opened 2022-02-08
docs: add instructions for opting out of Disable Resources feature

lizzthabet:
Add a short section to Disable Resources docs with instructions to disable the feature. We can remove these instructions when we have a better UI option.


hyu (approved):
Nice!

PR #1062, opened 2022-02-11
Fix incorrect helm_remote link

kalmant:
The current link for `helm_resource` in the docs results in a 404. I believe the author wanted to link the extension on GitHub.


nicks (approved):
Thanks @kalmant ! LGTM

(though i think @nicksieger is working on autogenerating there)

PR #1094, opened 2022-04-27
docs: add note about ignoring `.git` directory workaround

lizzthabet:
Based on feedback that's recently come up from [#2169](https://github.com/tilt-dev/tilt/issues/2169), I've added notes about Tilt's ignore and watch rules so it's easier to find when someone is looking at `docker_build()` documentation.

Thanks to everyone for the feedback!


nicksieger (approved):

docs/file_changes.md
```diff
@@ -64,11 +64,15 @@ Whenever a file under that path changes, Tilt will re-run the specified scripts.
 
 ### .git
 
-Tilt will always ignore changes under the `.git` directory.
+Tilt will always ignore changes under the `.git` directory, since watching the
+directory may trigger unintended builds and impact performance.
```
electrofelix, line 68: Can you expand on why tilt can't just ignore changes to the .git directory without removing it from the docker build context?

I suspect also that it might help users to flag that .git is handled specially in the api reference under docker_build, as unless they start looking at excluding files for to reduce rebuilds, less likely to look into the Guide to File Changes when taking tilt for a test drive, and this is possibly a sufficient surprise when the expectation is that `docker_build` => `docker build ...`

README.md
```diff
@@ -30,12 +30,14 @@ When you're finished, merge to master and Netlify will auto-deploy the site.
...
-You generally shouldn't need to run `make api` by hand. The Tiltfile
-will do it automatically whenever api/api.py changes.
+We write the API docs as docstrings in Python files in the main
+[Tilt repo](https://github.com/tilt-dev/tilt). Then, during the
+release pipeline (`release-update-tilt-docs.sh`), the files are
+generated and updated here (under `api/api`), Sphinx generates HTML
+fragments (`make api`), and then Jekyll serves that HTML fragment.
+
+The autogenerated API Python files include instructions on
+how to edit the source content.
```
nicksieger, line 40: 🎉

docs/file_changes.md
lizzthabet: yeah, that's a great point! i've added a note and link about file ignores to the api docs in [#5748](https://github.com/tilt-dev/tilt/pull/5748). those changes will go out in the next tilt release today. 

to answer your question, i think this is a solution we should talk about in #2169. when you configure a `docker_build` with `ignore` or `only`, tilt excludes those files from both the docker context and from file watching. similarly, when you define a context, tilt will watch all of those files. (that's to say that there isn't a way to configure the build context and file watching separately with tilt unless you use `custom_build`.) the changes that nick mentioned in [#5708](https://github.com/tilt-dev/tilt/pull/5708) should make it easier to implement an option to include `.git` in the context without watching the directory. 

thanks again for taking a look at these changes! i hope this adds some clarity.

PR #1103, opened 2022-05-11
docs: update for offline Snapshots

hyu:
Instead of deleting these pages outright, I'm editing them heavily to reflect our new approach.


milas (approved):
I like the new snapshots page! I do think it might be better to delete the Tilt Cloud pages entirely because I don't know that they have any value anymore (plus have been unlinked for some time as-is)

milas:
See also #1104 for blog post / announcement

hyu:
> I like the new snapshots page! I do think it might be better to delete the Tilt Cloud pages entirely because I don't know that they have any value anymore (plus have been unlinked for some time as-is)

Twist my arm! But yeah, I think you're right, they don't really add a ton. While users could still get to the pages via search and other links (e.g., old blog post), it was just one or two hits a week. I think the more discoverable Snapshots page with its Tilt Cloud reference should explain the situation.

PR #1136, opened 2023-06-02
Fix link for Why tilt uses immutabe tags

dnephin:
I noticed this was broken here: https://docs.tilt.dev/api.html#api.custom_build

Looking at some of the other links in this section I think this might be the fix.


dnephin:
Does not appear to be fixed, I'm not sure why.

nicks (approved):
lgtm

thanks! the docs will get auto-updated on the next release

PR #1137, opened 2023-06-12
Update helm.md to help users avoid a race condition

chase-miller:
Added `resource_deps` argument. Without this, this tilt resource will fail on the first run (i.e. where the repo has not yet been added to helm). Also updated to demonstrate how to sync k8s descriptors against a helm_resource that would otherwise fail to apply on first execution.

### DIFF
--- a/docs/helm.md
+++ b/docs/helm.md
@@ -32,16 +32,7 @@ chart](https://artifacthub.io/packages/helm/bitnami/mysql):
 
 load('ext://helm_resource', 'helm_resource', 'helm_repo')
 helm_repo('csi-secrets-store-provider-azure', 'https://azure.github.io/secrets-store-csi-driver-provider-azure/charts')
-helm_resource(
-  'csi',
-  'csi-secrets-store-provider-azure/csi-secrets-store-provider-azure',
-  flags=[
-    '--set', 'secrets-store-csi-driver.syncSecret.enabled=true',
-    '--set', 'secrets-store-csi-driver.enableSecretRotation=true'
-  ], 
-  namespace='kube-system',
-  resource_deps=['csi-secrets-store-provider-azure']
-)
+helm_resource('csi', 'csi-secrets-store-provider-azure/csi-secrets-store-provider-azure', resource_deps=['csi-secrets-store-provider-azure'])
 
 k8s_resource(
   objects=['ingestion-secretproviderclass:secretproviderclass'],
--- a/docs/helm.md
+++ b/docs/helm.md
@@ -30,14 +30,26 @@ chart](https://artifacthub.io/packages/helm/bitnami/mysql):
 ```
 # Tiltfile
 
+helm_repo('bitnami', 'https://charts.bitnami.com/bitnami')
+helm_resource('mysql', 'bitnami/mysql', resource_deps=['bitnami'])
+```
+
+And one that deploys the [secrets-store-csi-driver-provider-azure `csi` chart](https://azure.github.io/secrets-store-csi-driver-provider-azure/docs/getting-started/installation/) on which a `SecretProviderClass` CRD relies:
+
+```
+# Tiltfile
+
 load('ext://helm_resource', 'helm_resource', 'helm_repo')
 helm_repo('csi-secrets-store-provider-azure', 'https://azure.github.io/secrets-store-csi-driver-provider-azure/charts')
-helm_resource('csi', 'csi-secrets-store-provider-azure/csi-secrets-store-provider-azure', resource_deps=['csi-secrets-store-provider-azure'])
+helm_resource(
+  'csi',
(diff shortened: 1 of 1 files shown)


docs/helm.md
```diff
@@ -31,8 +31,14 @@ chart](https://artifacthub.io/packages/helm/bitnami/mysql):
 # Tiltfile
 
 load('ext://helm_resource', 'helm_resource', 'helm_repo')
-helm_repo('bitnami', 'https://charts.bitnami.com/bitnami')
-helm_resource('mysql', 'bitnami/mysql')
+helm_repo('csi-secrets-store-provider-azure', 'https://azure.github.io/secrets-store-csi-driver-provider-azure/charts')
```
nicks, line 34: can you change it back to the mysql example? trying to keep these examples simple

chase-miller: sure. can you think of a real-world example of a k8s resource that relies on this helm release being installed before it can be applied to the cluster? I know very little about this helm chart. maybe something like the following?

```
load('ext://helm_resource', 'helm_resource', 'helm_repo')
helm_repo('bitnami', 'https://charts.bitnami.com/bitnami')
helm_resource('mysql', 'bitnami/mysql', resource_deps=['bitnami'])

k8s_resource(
  objects=['my-db:mysql'],
  new_name='my-db',
  resource_deps=['mysql'] # k8s will fail to apply resources on a cluster where mysql has just been provisioned (e.g. a brand new one). This tells it to wait until the mysql helm_resource has finished before applying to the cluster.
)
```

I'm hoping that the documentation can convey bridge two current gaps:
1. `helm_resource` failing to execute because the helm repo hasn't yet been added (e.g. on the first run).
2. `k8s_yaml` failing to execute because kubectl doesn't yet recognize a kind (e.g. on the first run).

nicks: also, have you tested that this example even works? i don't think the `helm_resource` chart creates a k8s object that you can re-assign until after it deploys.

chase-miller: > how about we break it into two examples: one without crds (that just uses helm_repo and helm_resource, like the mysql one) and one with crds (that needs the extra k8s resource)

Love it. I'll update the PR to do this.

> also, have you tested that this example even works? i don't think the helm_resource chart creates a k8s object that you can re-assign until after it deploys.

Yep, it works. This is what we're using in our tilt chart (which is executed in a build pipeline via `tilt ci` against a fresh local kind cluster created using `ctlptl`). Without the `k8s_resource`, tilt fails running against a fresh cluster with the error below. Including the `k8s_resource` results in a successful tilt execution. I'm assuming that the `resource_deps` of the `k8s_resource` instructs tilt to wait until the `helm_resource` finishes before applying the SecretProviderClass.

```
Build Failed: kubernetes apply: error mapping secrets-store.csi.x-k8s.io/SecretProviderClass: no matches for kind "SecretProviderClass" in version "secrets-store.csi.x-k8s.io/v1"
```

chase-miller: I just pushed a commit that includes a CRD example. I used a csi example since I've tested this as an example (and know that `k8s_resource` is needed). I can change it to a prometheus example, but I'm not super confident how that would look.

nicks: re: "I used a csi example since I've tested this and know that the k8s_resource block is needed."

I tried your example in this PR:
https://github.com/tilt-dev/tilt-extensions/pull/508

I got this error:

```
No object identified by the fragment "my-secretproviderclass:secretproviderclass" could be found.
```

you can see the error here:
https://app.circleci.com/pipelines/github/tilt-dev/tilt-extensions/1393/workflows/3f7f132f-59de-43b8-9c89-3e268a8f8c8d/jobs/3275

what am i missing?

chase-miller: The example assumes a tilt resource that applies to k8s a descriptor with the name `my-secretproviderclass` and kind `secretproviderclass`. Here's a working example (forked from your PR): https://github.com/tilt-dev/tilt-extensions/pull/510/files. We use kustomize (our line is `k8s_yaml(kustomize('Deploy/kubernetes/local'))` instead of the added line in the PR).

nicks: in your example, there's an extra yaml file that the `k8s_resource` is selecting; it has nothing to do with the helm_resource. Not sure this belongs in the helm docs.

chase-miller: Without including something in the docs to tell a user to add a `k8s_resource` line, leveraging `helm_resource` fails on a clean cluster (or in a ci/cd pipeline) in the case where the resources helm applies are needed (like with a CRD). 

Similarly but separate, `helm_resource` will fail on a fresh environment without adding the necessary `resource_deps` as this PR demonstrates.

How will a user know what to do if it's not in the docs?

nicks:
added the resource_deps to the existing example.

i don't think the other example you added makes sense in a general doc; it depends on a lot of the internals of SecretProviderClass

PR #1138, opened 2023-06-27
deploy: update docs

nicks:


nicks:
unfortunately, hgrep started failing in the release pipeline so i moved this script over to pup for html processing https://github.com/ericchiang/pup

nicks:
Code blocks aren't formatting quite right :sob:

PR #1140, opened 2023-09-19
fixing kubernetes link from install.md

tom1mat:
the current link https://docs.docker.com/docker-for-mac/#kubernetes is not accurate because it seems like the get-started page has changed and doesn't have the kubernetes section anymore

Screenshot 2023-09-18 at 22 44 28

doing some googling it seems that this (https://docs.docker.com/desktop/kubernetes/) is the correct page


nicks (approved):
lgtm! thanks!

PR #1156, opened 2024-09-27
docs: update 'GitHub' brand name

hsnabszhdn:
It's a common mistake: The actual brand is "Git**H**ub" as opposed to "Git**h**ub".

There was references in `/blog/_posts/*` too, but I thought they were water under the bridge and no point in changing them, but happy to apply to them as well if needed.


nicks (approved):
lgtm!

PR #1157, opened 2024-09-27
docs: fix 'no-op' term

hsnabszhdn:
This seems to be referring to [Wikipedia: NOP](https://en.wikipedia.org/wiki/NOP_(code))

> In computer science, a NOP, **no-op**, or NOOP (pronounced "no op"; short for no operation) is a machine language instruction and its assembly language mnemonic, programming language statement, or computer protocol **command that does nothing**.


nicks (approved):
lgtm!

PR #1158, opened 2024-09-27
docs: use inline code for 'snack'

hsnabszhdn:
It's referenced via inline code block in two places in the same paragraph. This one is inconsistent.

image


nicks (approved):
lgtm! thanks!

PR #1175, opened 2025-04-30
Updated documentation for using $EXPECTED_REF on Windows

asos-alexhaigh:
$EXPECTED_REF does not work on Windows, support was added with "%EXPECTED_REF%" in this PR
https://github.com/tilt-dev/tilt-extensions/pull/401

Documentation was never updated, meaning I had to Google an error to find out about this.


nicks:
Thanks for pointing out this issue!

However, I don't think the suggested fix is correct -- you'd typically use command_bat on windows machines, as nicksieger pointed out in the pr you referenced.

https://docs.tilt.dev/api.html#api.custom_build

PR #1180, opened 2025-06-04
docs: fix helm references in API reference

marcoderama:
Could also go with https://helm.sh/docs/helm/helm_template/#helm-template but it didn't seem worth it to me.


marcoderama:
With respect to the second change, there is no `--name` flag for `helm template`.  This is referencing the `NAME` argument.
```
Usage:
  helm template [NAME] [CHART] [flags]
```

nicks:
Hmmm...I don't think this change is right, I thought we autogenerated these docs from https://github.com/tilt-dev/tilt/blob/master/internal/tiltfile/api/__init__.py#L1

marcoderama:
Thanks for the pointer.  Fixed.  (I hope :-)

marcoderama:
Last push just added the requisite `Signed-off-by` to the commit message.

PR #1185, opened 2025-08-10
Fix homebrew install command to use default formula

jacobwhall:
Tilt can be installed using [this formula](https://github.com/Homebrew/homebrew-core/blob/main/Formula/t/tilt.rb) instead of the custom tap.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -108,7 +108,7 @@ You can also use these installers directly.
 ### Homebrew (macOS or Linux)
 
 ```bash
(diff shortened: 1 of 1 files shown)


jacobwhall:
Perhaps this line should be changed, too.

https://github.com/jacobwhall/tilt.build/blob/4b419cc2e98f64832a7a95dc7ae881f78cfdab67/docs/upgrade.md?plain=1#L35

nicks:
thanks!

PR #1199, opened 2026-04-21
Update install.md

textractor:
added info on installing Tilt with Homebrew for Mac OS


nicks:
this is already covered in the alternative installation instructions below

